PR #2, opened 2023-08-24
Fix `ssh` rules

AleksandarSavchev:
This PR fixes rules `242393` and `242394`.

Rule `242393` when passed always reported with `SSH daemon inactive (or could not be probed)`, now it should report with `SSH daemon service not installed` when the `systemctl is-active sshd` command outputs `inactive`.

Rule `242394` returned errored status in cases it should have passed. Rule `242394` now correctly passes.

Fixes #

**Release note**:

```bugfix user
A bug causing rule `242393` to pass with wrong message was fixed.
```
```bugfix user
A bug causing rule `242394` to error when it should pass was fixed.
```


dimityrmirchev (approved):
/lgtm

PR #3, opened 2023-08-24
Improve `ETCD` peer options rules

AleksandarSavchev:
This PR improves `ETCD` peer options rules `242380`, `242426`, `242432` and `242433` by skipping them when ETCD runs as a single instance, since they are not used in that case.

Fixes #

**Release note**:

```feature user
`ETCD` peer options rules `242380`, `242426`, `242432` and `242433` are now skipped when `ETCD` runs as a single instance.
```


dimityrmirchev (approved):
/lgtm

PR #7, opened 2023-08-31
Export utils

dimityrmirchev:
This PR exposes some of the utility functions so that they can be used in rule implementations outside of this repository.

Fixes #

**Release note**:

```feature user

```


pkg/kubernetes/utils/utils.go
```diff
@@ -0,0 +1,393 @@
...
+
+		if len(nodeList.Continue) == 0 {
+			return nodes, nil
+		}
+	}
+}
+
+// GetWorkers return all workers for a given namespace, or all namespaces if it's set to empty string "".
+// It retrieves workers by portions set by limit.
+func GetWorkers(ctx context.Context, c client.Client, namespace string, limit int64) ([]extensionsv1alpha1.Worker, error) {
```
AleksandarSavchev, line 90: Is `pkg/kubernetes/utils` the right place for `GetWorkers()` considering that the `Worker` resource is Gardener specific? I think it should stay in `internal`. WDYT?

AleksandarSavchev (commented):
Nice PR! Only 1 minor suggestion.

pkg/kubernetes/utils/utils.go
dimityrmirchev: Yes, good point. I will find a better place for this function.

AleksandarSavchev (approved):
/lgtm

PR #12, opened 2023-09-12
Use label selector instead of name prefix for rule `242414`

AleksandarSavchev:
This PR changes rule `242414` to use pod and namespace label selectors for accepting pods instead of pod name and namespace prefixes.

Fixes #

**Release note**:

```other user
Selecting accepted pods for rule `242414` in the config file has been changed to use pod and namespace label selectors instead of name prefixes.
```

### DIFF
--- a/pkg/provider/gardener/internal/utils/utils.go
+++ b/pkg/provider/gardener/internal/utils/utils.go
@@ -105,6 +105,21 @@ func Subset(s1, s2 []string) bool {
 	return true
 }
 
+// MatchLabels checks if all m2 keys and values are present in m1. If m1 or m2 is nil returns false.
+func MatchLabels(a, b map[string]string) bool {
+	if a == nil || b == nil {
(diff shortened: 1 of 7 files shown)


pkg/provider/gardener/internal/utils/utils.go
```diff
@@ -105,6 +105,21 @@ func Subset(s1, s2 []string) bool {
 	return true
 }
 
+// MatchLabels checks if all m2 keys and values are present in m1. If m1 or m2 is nil returns false.
+func MatchLabels(a, b map[string]string) bool {
```
dimityrmirchev, line 109: a and b should be renamed to m1 and m2 to reflect the comment

dimityrmirchev (changes requested):
Just one comment from my side

dimityrmirchev (approved):
/lgtm

PR #39, opened 2023-10-03
`pod-files` checks only files with paths part of the `volumeMounts` of the container

AleksandarSavchev:

Fixes #

**Release note**:

```other user
DISA Kubernetes STIGS `pod-files` rule now checks only files with paths part of the `volumeMounts` for the specific container. It also excludes directories of no interest like `/var/log/journal`.
```

### DIFF
--- a/pkg/provider/gardener/ruleset/disak8sstig/v1r10/pod_files_permissions_and_owner.go
+++ b/pkg/provider/gardener/ruleset/disak8sstig/v1r10/pod_files_permissions_and_owner.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
+	"slices"
 	"strings"
 
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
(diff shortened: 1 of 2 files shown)


pkg/provider/gardener/ruleset/disak8sstig/v1r10/pod_files_permissions_and_owner.go
```diff
@@ -274,3 +283,44 @@ func (r *RulePodFiles) checkContainerd(
 
 	return checkResults
 }
+
+func (r *RulePodFiles) isMountRequiredByContainer(destination, containerName string, pod *corev1.Pod) bool {
+	for _, container := range pod.Spec.Containers {
+		if container.Name == containerName {
```
dimityrmirchev, line 289: We can reduce nesting if we continue when the names do not match like in the function below.

dimityrmirchev (changes requested):
Just one minor suggestion

dimityrmirchev (approved):
/lgtm

PR #48, opened 2023-10-10
`pod-files` rule allow user and group ID `65532`

AleksandarSavchev:

Fixes #

**Release note**:

```feature user
DISA Kubernetes STIGs `pod-files` rule now passes files with owner and/or group ID `65532`.
```

### DIFF
--- a/pkg/provider/gardener/ruleset/disak8sstig/v1r10/pod_files_permissions_and_owner.go
+++ b/pkg/provider/gardener/ruleset/disak8sstig/v1r10/pod_files_permissions_and_owner.go
@@ -227,8 +227,8 @@ func (r *RulePodFiles) checkContainerd(
 		value := pair[1]
 		if metav1.HasLabel(pod.ObjectMeta, key) && pod.Labels[key] == value {
 			delete(mandatoryComponents, component)
-			expectedFileOwnerUsers = []string{"0"}
-			expectedFileOwnerGroups = []string{"0", "65534"}
+			expectedFileOwnerUsers = []string{"0", "65532"}
+			expectedFileOwnerGroups = []string{"0", "65532", "65534"}
 		}
 	}
 


pkg/provider/gardener/ruleset/disak8sstig/v1r10/pod_files_permissions_and_owner.go
```diff
@@ -227,8 +227,8 @@ func (r *RulePodFiles) checkContainerd(
 		value := pair[1]
 		if metav1.HasLabel(pod.ObjectMeta, key) && pod.Labels[key] == value {
 			delete(mandatoryComponents, component)
-			expectedFileOwnerUsers = []string{"0"}
-			expectedFileOwnerGroups = []string{"0", "65534"}
+			expectedFileOwnerUsers = []string{"0", "65532"}
```
dimityrmirchev, line 230: Why do we set this in users but we do not set user id "65534"?

AleksandarSavchev: Gardener images use `distroless` `nonroot` user which has ID 65532 [ref](https://github.com/GoogleContainerTools/distroless/blob/main/base/base.bzl#L8). The `65534` ID was used only by the `etcd` image from [etcd-custom-image](https://github.com/gardener/etcd-custom-image) which is no longer in used and replaced with [etcd-wrapper](https://github.com/gardener/etcd-wrapper/blob/main/Dockerfile).

dimityrmirchev: Can you please add the reference link as a comment so we know where it is coming from?

dimityrmirchev (approved):
/lgtm

PR #49, opened 2023-10-10
`SimplePodExecutor` `Execute()` return `stderr` before error code

AleksandarSavchev:
When bash commands error they sometimes output an error message on stderr, exit with an exit code or do both. This PR changes the `SimplePodExecutor` `Execute` method to return the stderr first, if there is one, since it is more  descriptive.

Fixes #

**Release note**:

```feature user
NONE
```

### DIFF
--- a/pkg/kubernetes/pod/pod.go
+++ b/pkg/kubernetes/pod/pod.go
@@ -142,17 +142,17 @@ func (spe *SimplePodExecutor) Execute(ctx context.Context, command string, comma
 		Stderr: &stderr,
 		Tty:    false,
 	})
-	if err != nil {
-		return "", err
-	}
-
-	stderrByte, err := io.ReadAll(&stderr)
-	if err != nil {
-		return "", err
+	stderrByte, otherErr := io.ReadAll(&stderr)
+	if otherErr != nil {
+		return "", otherErr
 	}
 
 	if len(stderrByte) > 0 {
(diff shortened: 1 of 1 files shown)


pkg/kubernetes/pod/pod.go
```diff
@@ -142,17 +142,17 @@ func (spe *SimplePodExecutor) Execute(ctx context.Context, command string, comma
 		Stderr: &stderr,
 		Tty:    false,
 	})
-	if err != nil {
-		return "", err
+	stderrByte, otherErr := io.ReadAll(&stderr)
```
dimityrmirchev, line 145: We can do something in the likes of 

```golang
if err != nil && otherErr != nil {
	return "", errors.Join(err, otherErr)
} else if otherErr != nil {
	return "", otherErr
}

if err != nil && len(stderrByte) > 0 {
	return "", fmt.Errorf("err: %w, command %s %s stderr output: %s", err, command, commandArg, string(stderrByte))
} else if len(stderrByte) > 0 {
	return "", fmt.Errorf("command %s %s stderr output: %s", command, commandArg, string(stderrByte))
}

if err != nil {
	return "", fmt.Errorf("err: %w, command %s %s", err, command, commandArg)
}
```

dimityrmirchev (changes requested):

dimityrmirchev (approved):
/lgtm

PR #54, opened 2023-10-16
Add GetSingleAllocatableNodePerWorker function to internal utils

AleksandarSavchev:
Add `GetSingleAllocatableNodePerWorker()` to internal utils, which is an improved version of `GetSingleRunningNodePerWorker()`. `GetSingleAllocatableNodePerWorker()` also checks if the node has at least 1 allocatable spot.

Fixes #

**Release note**:

```feature user
NONE
```

### DIFF
--- a/pkg/provider/gardener/internal/utils/utils.go
+++ b/pkg/provider/gardener/internal/utils/utils.go
@@ -42,7 +42,7 @@ func GetWorkers(ctx context.Context, c client.Client, namespace string, limit in
 }
 
 // AllocatableNode contains a single Node and whether it is in allocatable or not.
-// A allocatable node is a node that is in ready state and has at least 1 allocatable spot.
+// An allocatable node is a node that is in ready state and has at least 1 allocatable spot.
 type AllocatableNode struct {
 	Node        *corev1.Node
 	Allocatable bool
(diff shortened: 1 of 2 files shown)


pkg/provider/gardener/internal/utils/utils.go
```diff
@@ -41,6 +41,46 @@ func GetWorkers(ctx context.Context, c client.Client, namespace string, limit in
...
+					maxAllocatablePods < nodesAllocatablePods[node.Name] {
+					maxAllocatablePods = nodesAllocatablePods[node.Name]
+					node := node
+					allocatableNode = &node
+				}
+			}
+
+			result[workerGroup.Name] = AllocatableNode{
+				Node:        allocatableNode,
+				Allocatable: allocatableNode != nil, // assign a nil not allocatable entry for the worker group if we could not find a allocatable node
```
dimityrmirchev, line 76:
```suggestion
				Allocatable: allocatableNode != nil, // assign a nil not allocatable entry for the worker group if we could not find an allocatable node
```

pkg/provider/gardener/internal/utils/utils.go
```diff
@@ -41,6 +41,46 @@ func GetWorkers(ctx context.Context, c client.Client, namespace string, limit in
...
+
+			var allocatableNode *corev1.Node
+			maxAllocatablePods := 0
+			for _, node := range nodes {
+				// find a allocatable node with the most allocatable spots for pods
+				if node.ObjectMeta.Labels[v1beta1constants.LabelWorkerPool] == workerGroup.Name && kubeutils.NodeReadyStatus(node) &&
+					maxAllocatablePods < nodesAllocatablePods[node.Name] {
+					maxAllocatablePods = nodesAllocatablePods[node.Name]
+					node := node
+					allocatableNode = &node
```
dimityrmirchev, line 70: Why do we make a copy here? Maybe we can change the name of the copy to something other than `node` so it differs from the loop `node` variable

dimityrmirchev (changes requested):

dimityrmirchev (approved):
/lgtm

PR #56, opened 2023-10-16
[Kubernetes STIG v1r10] 242442: check only Gardener managed pods in shoot

AleksandarSavchev:

Fixes #

**Release note**:

```other user
DISA Kubernetes STIGs `242442` rule no longer checks shoot pods that are not managed by Gardener.
```


dimityrmirchev:
Why is this change needed?

AleksandarSavchev:
> Why is this change needed?

Shoot owners can create their own pods on the shoot, which happen to use the same images as pods on the control-plane or on shoot's `kube-system` namespace. I think this rule should only take into consideration `Gardener` workload.

dimityrmirchev (approved):
/lgtm

PR #57, opened 2023-10-17
[Kubernetes STIG v1r10] 242387:  create pods only on nodes with allocatable spot

AleksandarSavchev:
Currently `Kubelet` rules create a pod on the first `Ready` node found for a worker group. The first node might not have allocatable spots and the rule would return an error. This PR fixes the problem for rule 242387 by always using a node with allocatable spots or if there is none such node for a worker group the rule returns a warning.

If this PR is approved PRs with the same changes for the other `Kubelet` rules will be opened.

Fixes #

**Release note**:

```feature user
NONE
```

### DIFF
--- a/pkg/provider/gardener/ruleset/disak8sstig/v1r10/242387.go
+++ b/pkg/provider/gardener/ruleset/disak8sstig/v1r10/242387.go
@@ -124,7 +124,7 @@ func (r *Rule242387) Run(ctx context.Context) (rule.RuleResult, error) {
 func (r *Rule242387) checkWorkerGroup(ctx context.Context, workerGroup string, node utils.AllocatableNode, privPodImage string) rule.CheckResult {
 	target := gardener.NewTarget("cluster", "seed", "kind", "workerGroup", "name", workerGroup)
 	if !node.Allocatable {
-		return rule.WarningCheckResult("There are no nodes with at least 1 allocatable spot for worker group.", target)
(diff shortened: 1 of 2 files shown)


pkg/provider/gardener/ruleset/disak8sstig/v1r10/242387.go
```diff
@@ -114,10 +121,10 @@ func (r *Rule242387) Run(ctx context.Context) (rule.RuleResult, error) {
 	}, nil
 }
 
-func (r *Rule242387) checkWorkerGroup(ctx context.Context, workerGroup string, node utils.ReadyNode, privPodImage string) rule.CheckResult {
+func (r *Rule242387) checkWorkerGroup(ctx context.Context, workerGroup string, node utils.AllocatableNode, privPodImage string) rule.CheckResult {
 	target := gardener.NewTarget("cluster", "seed", "kind", "workerGroup", "name", workerGroup)
-	if !node.Ready {
-		return rule.WarningCheckResult("There are no nodes in Ready state for worker group.", target)
+	if !node.Allocatable {
+		return rule.WarningCheckResult("There are no nodes with at least 1 allocatable spot for worker group.", target)
```
dimityrmirchev, line 127:
```suggestion
		return rule.WarningCheckResult("There are no ready nodes with at least 1 allocatable spot for worker group.", target)
```

dimityrmirchev (changes requested):
Just one minor recommendation

dimityrmirchev (approved):
/lgtm

PR #59, opened 2023-10-19
[Kubernetes STIG v1r10] Kubelet rules: create pods only on nodes with allocatable spots

AleksandarSavchev:
Create pods only on nodes with allocatable spots for all Kubelet rules, as in #57.

Fixes #

**Release note**:

```bugfix user
DISA Kubernetes STIGs Kubelet rules now creates `diki` pods only on nodes with free allocatable space.
```


dimityrmirchev (approved):
/lgtm

PR #60, opened 2023-10-24
[Kubernetes STIG v1r10] pod-files rule: add check for 242467 rule

AleksandarSavchev:
This PR improves the *.key files checks required from `242467` rule in `pod-files` rule. *.key files in mandatory components are required to have max privilege of `640`. We do not enforce `600` because k8s does not provide a way to easily change the owner of a file and containers are expected to run as nonroot.

Fixes #

**Release note**:

```bugfix user
DISA Kubernetes STIGs `pod-files` rule now expects `0640` permission setting for `*.key` files of mandatory components. This change improves the `242467` rule which requires `0600` permissions for such files. `0600` is not enforced since k8s does not provide an easy way to change the owner of a file and containers are expected to run as nonroot. 
```

### DIFF
--- a/pkg/provider/gardener/ruleset/disak8sstig/v1r10/pod_files_permissions_and_owner.go
+++ b/pkg/provider/gardener/ruleset/disak8sstig/v1r10/pod_files_permissions_and_owner.go
@@ -322,8 +322,10 @@ func (r *RulePodFiles) checkContainerd(
 
 				if isMandatoryComponent && strings.HasSuffix(strings.Join(statSlice[3:], " "), ".key") {
 					// rule 242467
+					// Gardener control plane components run as `nonroot` user `65532`, since we can change the group owener but
+					// cannot easily change the user owner of secret files we do not check for `0600` permission but instead for `0640`.
 					checkResults = append(checkResults, utils.MatchFilePermissionsAndOwnersCases(statSlice[0], statSlice[1], statSlice[2], strings.Join(statSlice[3:], " "),
-						"600", expectedFileOwnerUsers, expectedFileOwnerGroups, fileTarget)...)
+						"640", expectedFileOwnerUsers, expectedFileOwnerGroups, fileTarget)...)
 					continue
 				}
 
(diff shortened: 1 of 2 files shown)


pkg/provider/gardener/ruleset/disak8sstig/v1r10/pod_files_permissions_and_owner.go
```diff
@@ -53,6 +53,13 @@ type ExpectedFileOwner struct {
 	Groups []string `yaml:"groups"`
 }
 
+type component struct {
+	Name  string
+	Label string
+	Value string
+	Num   int
```
dimityrmirchev, line 60: Let's remove the Num field and again use a `map[string]component`

pkg/provider/gardener/ruleset/disak8sstig/v1r10/pod_files_permissions_and_owner_test.go
```diff
@@ -49,6 +49,8 @@ var _ = Describe("#RulePodFiles", func() {
 ]`
 		compliantStats = `600 0 0 /destination/file1.txt
 644 0 65532 /destination/bar/file2.txt`
+		keyFileStats = `600 0 0 /destination/file1.key
```
dimityrmirchev, line 52:
```suggestion
		keyFileStats = `640 0 0 /destination/file1.key
```

dimityrmirchev (changes requested):
Some minor requests from my side

pkg/provider/gardener/ruleset/disak8sstig/v1r10/pod_files_permissions_and_owner.go
AleksandarSavchev: We cannot use the old method of removing a found component from the map, since a mandatory component can have more than 1 pod and we want to make the `key` file checks for every single pod of that component. Using the old logic we would know only for the first pod of the component that we need to check the `key` files.

I suggest I change the `Num int` filed to `Found bool`

dimityrmirchev: Ok then. Lets change "Num" to "found"

dimityrmirchev (approved):
/lgtm

PR #70, opened 2023-11-28
Change the type of rule.Target

dimityrmirchev:
Changes the type of `rule.Target` to a `map[string]string` which I think will be easier and simpler to work with instead of `any`.

Fixes #

**Release note**:

```feature user
NONE
```


pkg/provider/gardener/targets_test.go
AleksandarSavchev, line 1: This file is no longer needed, since the test object has been moved out of this package and the test logic has been added there as well `pkg/rule/rule_test.go`.

AleksandarSavchev (changes requested):
Looks good, just a single nit.

pkg/provider/gardener/targets_test.go
dimityrmirchev: Yo are right. Thanks!

AleksandarSavchev (approved):
/lgtm

Issue #68, opened 2023-11-28
Implement a VirtualGarden provider


dimityrmirchev:
**What would you like to be added**:
A [VirtualGarden](https://gardener.cloud/docs/glossary/) provider that will implement a subset of the DISA K8s STIG rules. A suggestion for the subset:

- [x] 242376 The Kubernetes Controller Manager must use TLS 1.2, at a minimum, to protect the confidentiality of sensitive data during electronic dissemination. https://github.com/gardener/diki/pull/71
- [x] 242378 The Kubernetes API Server must use TLS 1.2, at a minimum, to protect the confidentiality of sensitive data during electronic dissemination. https://github.com/gardener/diki/pull/73
- [x] 242379 The Kubernetes etcd must use TLS to protect the confidentiality of sensitive data during electronic dissemination. https://github.com/gardener/diki/pull/75
- [x] 242380 The Kubernetes etcd must use TLS to protect the confidentiality of sensitive data during electronic dissemination. https://github.com/gardener/diki/pull/75
- [x] 242381 The Kubernetes Controller Manager must create unique service accounts for each work payload. https://github.com/gardener/diki/pull/76
- [x] 242382 The Kubernetes API Server must enable Node,RBAC as the authorization mode. https://github.com/gardener/diki/pull/76
- [x] 242383 User-managed resources must be created in dedicated namespaces. https://github.com/gardener/diki/pull/76
- [x] 242384 The Kubernetes Scheduler must have secure binding. https://github.com/gardener/diki/pull/76
- [x] 242385 The Kubernetes Controller Manager must have secure binding. https://github.com/gardener/diki/pull/77
- [x] 242386 The Kubernetes API server must have the insecure port flag disabled. https://github.com/gardener/diki/pull/77
- [x] 242388 The Kubernetes API server must have the insecure bind address not set. https://github.com/gardener/diki/pull/77
- [x] 242389 The Kubernetes API server must have the secure port set. https://github.com/gardener/diki/pull/78
- [x] 242390 The Kubernetes API server must have anonymous authentication disabled. https://github.com/gardener/diki/pull/78
- [x] 242400 The Kubernetes API server must have Alpha APIs disabled. https://github.com/gardener/diki/pull/80
- [x] 242402 The Kubernetes API Server must have an audit log path set. https://github.com/gardener/diki/pull/80
[...]

PR #71, opened 2023-11-29
Virtual Garden Provider implementation

dimityrmirchev:
This PR implements the basics of a virtual garden provider. It also shares the `242376` rule logic between providers. Par of https://github.com/gardener/diki/issues/68

Fixes #

We should also extend the new version script https://github.com/gardener/diki/blob/main/hack/gen-new-disak8sstig-version.sh but that can be done in another PR.

**Release note**:

```feature developer
Diki now has a basic implementation of a virtual garden provider.
```


pkg/provider/builder/virtualgarden.go
```diff
@@ -0,0 +1,51 @@
...
+
+// VirtualGardenProviderFromConfig retuns a Provider from a [ProviderConfig].
+func VirtualGardenProviderFromConfig(conf config.ProviderConfig) (provider.Provider, error) {
+	p, err := virtualgarden.FromGenericConfig(conf)
+	if err != nil {
+		return nil, err
+	}
+
+	setConfigDefaults(p.GardenConfig)
+	setConfigDefaults(p.RuntimeConfig)
```
AleksandarSavchev, line 26: I think the `setConfigDefaults()` function should be moved to a new utils file. ex. `pkg/provider/builder/utils.go`

pkg/shared/disak8sstig/v1r11/ids.go
AleksandarSavchev, line 1: We should remove the [pkg/provider/gardener/ruleset/disak8sstig/v1r11/options.go](https://github.com/gardener/diki/blob/main/pkg/provider/gardener/ruleset/disak8sstig/v1r11/options.go) file and use the constants from this file. We can do this in a new PR.

pkg/provider/virtualgarden/provider.go
AleksandarSavchev, line 1: You can add a unit test file as we have for the `gardener` provider: [ref](https://github.com/gardener/diki/blob/main/pkg/provider/gardener/provider_test.go)

example/config/virtualgarden.yaml
```diff
@@ -0,0 +1,16 @@
...
+    gardenKubeconfigPath: /tmp/garden.config  # path to garden cluster admin kubeconfig
+    runtimeKubeconfigPath: /tmp/runtime.config  # path to runtime cluster admin kubeconfig
+  rulesets:
+  - id: disa-kubernetes-stig
+    name: DISA Kubernetes Security Technical Implementation Guide
+    version: v1r11
+    ruleOptions: []
+output:
+  path: /tmp/test-output.json          #  optional, path to summary json report
+  minStatus: Passed
```
AleksandarSavchev, line 16:
```suggestion
providers:                # contains information about known providers
- id: virtualgarden       # unique provider identifier
  name: "Virtual Garden"  # user friendly name of the provider
  metadata:
    foo: bar
  args:
    gardenKubeconfigPath: /tmp/garden.config    # path to garden cluster admin kubeconfig
    runtimeKubeconfigPath: /tmp/runtime.config  # path to runtime cluster admin kubeconfig
  rulesets:
  - id: disa-kubernetes-stig
    name: DISA Kubernetes Security Technical Implementation Guide
    version: v1r11
    ruleOptions: []
output:
  path: /tmp/test-output.json  #  optional, path to summary json report
  minStatus: Passed
```

AleksandarSavchev (changes requested):
I think we should move the `shared` dir into `pkg/provider/shared`. WDYT?

pkg/provider/builder/virtualgarden.go
dimityrmirchev: I think that there is no need for another file right now because it is a single function

dimityrmirchev:
> I think we should move the shared dir into pkg/provider/shared. WDYT?

I addd a `ruleset` folder in between. I think that should be fine for now

AleksandarSavchev (approved):
/lgtm

PR #80, opened 2023-12-05
[DISA K8s STIG] Share, Skip and implement most of the virtual garden provider rules

dimityrmirchev:
Part of #68 Some of the rules are skipped with `NotImplemented` status. This is primary because we want to refactor these rules and separate them from the `node-files` and `pod-files` combined rules which right now are not reusable across providers.

Fixes #

**Release note**:

```feature user
NONE
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -25,6 +25,10 @@ issues:
       linters:
         - revive
       text: "exported: exported"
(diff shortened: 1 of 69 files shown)


AleksandarSavchev (approved):
Great PR! \lgtm

PR #83, opened 2024-01-02
Add MatchFileOwnersCases() to internal utils

AleksandarSavchev:
Added new `MatchFileOwnersCases()` function to internal utils that will be used to check file owners in the new split rules from `node-files` and `pod-files`. The old `MatchFilePermissionsAndOwnersCases` will not be used in the new rules and will be removed with the removal of the `node-file` and `pod-file` rules.

Fixes #

**Release note**:

```feature user
NONE
```

### DIFF
--- a/pkg/provider/gardener/internal/utils/utils.go
+++ b/pkg/provider/gardener/internal/utils/utils.go
@@ -157,7 +157,39 @@ func ExceedFilePermissions(filePermissions, filePermissionsMax string) (bool, er
 	return fileModePermission&^fileModePermissionsMax != 0, nil
 }
 
-// MatchFilePermissionsAndOwnersCases returns []rule.CheckResult for a given file and its permissions and owners  for a select expected values.
+// MatchFileOwnersCases returns []rule.CheckResult for a given file and its owners for a select expected values.
+func MatchFileOwnersCases(
+	fileOwnerUser,
+	fileOwnerGroup,
+	fileName string,
+	expectedFileOwnerUsers,
+	expectedFileOwnerGroups []string,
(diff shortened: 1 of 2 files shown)


pkg/provider/gardener/internal/utils/utils.go
```diff
@@ -157,7 +157,39 @@ func ExceedFilePermissions(filePermissions, filePermissionsMax string) (bool, er
...
+func MatchFileOwnersCases(
+	fileOwnerUser,
+	fileOwnerGroup,
+	fileName string,
+	expectedFileOwnerUsers,
+	expectedFileOwnerGroups []string,
+	target rule.Target,
+) []rule.CheckResult {
+	checkResults := []rule.CheckResult{}
+	if len(expectedFileOwnerUsers) > 0 {
```
dimityrmirchev, line 170: When do we expect these to be of len == 0? Shouldn't we always expect an owner 🤔 ?

AleksandarSavchev: I wanted to add a way to pass for every owner (user or group) and decided to pass when the expected owner slice was empty. There is no specific example, currently, that would use this case.
The opposite approach of failing when the expected slice is empty didn't make much sense, since there should always be a pass case.

dimityrmirchev: The pass case should be when expected equals actual. I would vote to remove both checks for `expectedFileOwnerUsers` and `expectedFileOwnerGroups` if there is no current use case.

dimityrmirchev (approved):
Thanks

/lgtm

PR #98, opened 2024-01-11
[DISA K8s STIG] Use diki-ops image for pod executor

AleksandarSavchev:
This PR replaces the use of the `ops-toolbelt` image with `diki-ops` for DISA K8s STIG V1R11 ruleset. The `diki-ops` is considerably lighter and contains only the needed tools for `diki`.

Fixes #

**Release note**:

```feature user
Diki now uses a lighter image for pod executors in DISA K8s STIG V1R11 ruleset
```

### DIFF
--- a/pkg/provider/gardener/ruleset/disak8sstig/v1r11/242387.go
+++ b/pkg/provider/gardener/ruleset/disak8sstig/v1r11/242387.go
@@ -61,9 +61,9 @@ func (r *Rule242387) Run(ctx context.Context) (rule.RuleResult, error) {
 		return rule.SingleCheckResult(r, rule.ErroredCheckResult(err.Error(), rule.NewTarget("cluster", "seed", "kind", "workerList"))), nil
 	}
 
-	image, err := imagevector.ImageVector().FindImage(ruleset.OpsToolbeltImageName)
+	image, err := imagevector.ImageVector().FindImage(ruleset.DikiOpsImageName)
 	if err != nil {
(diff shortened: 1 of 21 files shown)


pkg/provider/gardener/ruleset/disak8sstig/v1r11/242387.go
```diff
@@ -61,9 +62,15 @@ func (r *Rule242387) Run(ctx context.Context) (rule.RuleResult, error) {
...
 	}
 
-	image, err := imagevector.ImageVector().FindImage(ruleset.OpsToolbeltImageName)
+	image, err := imagevector.ImageVector().FindImage(ruleset.DikiOpsImageName)
 	if err != nil {
-		return rule.RuleResult{}, fmt.Errorf("failed to find image version for %s: %w", ruleset.OpsToolbeltImageName, err)
+		return rule.RuleResult{}, fmt.Errorf("failed to find image version for %s: %w", ruleset.DikiOpsImageName, err)
+	}
+
+	// check if tag is not present and use the controller's version if that is the case
```
dimityrmirchev, line 70:
```suggestion
	// check if tag is not present and use diki's version as a default
```

Same for other occurrences. When it becomes available in this repo we can use https://github.com/gardener/gardener/blob/5fc5009be3f3dc09ca08e991da786a3a4a9d0efa/pkg/utils/imagevector/imagevector.go#L381 instead of the current logic.

dimityrmirchev (changes requested):

dimityrmirchev (approved):
/lgtm

Issue #82, opened 2024-01-02
Split combined rules into single rules


AleksandarSavchev:
**What would you like to be added**:
DISA K8s STIG V1R11 rules `node-files` and `pod-files` should be split into single rules:

- [X] `node-files`
  - [X] 242406 The Kubernetes kubelet configuration file must be owned by root: #85 
  - [X] 242407 The Kubernetes KubeletConfiguration files must have file permissions set to 644 or more restrictive: #84 
  - [x] 242449 The Kubernetes Kubelet certificate authority file must have file permissions set to 644 or more restrictive: #86
  - [X] 242450 The Kubernetes Kubelet certificate authority must be owned by root: #149 
  - [x] 242451 The Kubernetes component PKI must be owned by root: #162 
  - [X] 242452 The Kubernetes kubelet config must have file permissions set to 644 or more restrictive: #89
  - [X] 242453 The Kubernetes kubelet config must be owned by root: #90
  - [X] 242466 The Kubernetes PKI CRT must have file permissions set to 644 or more restrictive: #152
  - [X] 242467 The Kubernetes PKI keys must have file permissions set to 600 or more restrictive: #160 
- [x] `pod-files`
  - [X] 242445 The Kubernetes component etcd must be owned by etcd https://github.com/gardener/diki/pull/119
  - [X] 242446 The Kubernetes conf files must be owned by root https://github.com/gardener/diki/pull/116
  - [X] 242447 The Kubernetes Kube Proxy must have file permissions set to 644 or more restrictive #134 
  - [x] 242448 The Kubernetes Kube Proxy must be owned by root #141
  - [X] 242459 The Kubernetes etcd must have file permissions set to 644 or more restrictive https://github.com/gardener/diki/pull/101
  - [X] 242466 The Kubernetes PKI CRT must have file permissions set to 644 or more restrictive https://github.com/gardener/diki/pull/117
  - [X] 242467 The Kubernetes PKI keys must have file permissions set to 600 or more restrictive https://github.com/gardener/diki/pull/108

PR #151, opened 2024-02-26
[DISA K8s STIG] Remove pod-files rule

AleksandarSavchev:
Since all the rules that `pod-files` checks are now implemented as stand alone rules (ref #82) there is not need for the `pod-files` rule.

Fixes #

**Release note**:

```feature user
NONE
```


dimityrmirchev (approved):
/lgtm

Issue #67, opened 2023-11-28
Implement a managed K8s provider


dimityrmirchev:
**What would you like to be added**:
A managed K8s provider that can implement a subset of the DISA K8s STIG rules. A suggestion for rules that can be implemented:

- [x] 242383 User-managed resources must be created in dedicated namespaces. #163 
- [X] 242387 The Kubernetes Kubelet must have the "readOnlyPort" flag disabled. #135 
- [x] 242391 The Kubernetes Kubelet must have anonymous authentication disabled. #136 
- [x] 242392 The Kubernetes kubelet must enable explicit authorization. #137 
- [x] 242393 Kubernetes Worker Nodes must not have sshd service running. #140 
- [X] 242394 Kubernetes Worker Nodes must not have the sshd service enabled. #148 
- [X] 242395 Kubernetes dashboard must not be enabled. (note: depends on the implementation) #131 
- [X] 242396 Kubernetes Kubectl cp command must give expected access and results. #155
- [x] 242397 The Kubernetes kubelet staticPodPath must not enable static pods. #138 
- [X] 242398 Kubernetes DynamicAuditing must not be enabled. **Option feature-gates.DynamicAuditing was removed in Kubernetes v1.19.** https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates-removed/
- [x] 242399 Kubernetes DynamicKubeletConfig must not be enabled. #142
- [x] 242400 The Kubernetes API server must have Alpha APIs disabled. For kube-proxy. #156 
- [x] 242404 Kubernetes Kubelet must deny hostname override. #150 
- [X] 242406 The Kubernetes KubeletConfiguration file must be owned by root. #85 
- [X] 242407 The Kubernetes KubeletConfiguration files must have file permissions set to 644 or more restrictive. #84
- [X] 242414 The Kubernetes cluster must use non-privileged host ports for user pods. #128 
- [X] 242415 Secrets in Kubernetes must not be stored as environment variables. https://github.com/gardener/diki/pull/123
- [X] 242417 Kubernetes must separate user functionality. #130
- [x] 242420 Kubernetes Kubelet must have the SSL Certificate Authority set. #139 
- [x] 242424 Kubernetes Kubelet must enable tlsPrivateKeyFile for client authentication to secure service. #143 
- [x] 242425 Kubernetes Kubelet must enable tlsCertFile for client authentication to secure service. #144 
- [x] 242434 Kubernetes Kubelet must enable kernel protection. #145
[...]

PR #153, opened 2024-02-29
[DISA K8s STIG] Implement rule 254801

AleksandarSavchev:
This PR improves DISA K8s STIGs rule 254801 for gardener by checking control plane components as well and implements the rule for managedk8s provider.

Part of #67

**Release note**:

```feature user
NONE
```


dimityrmirchev (changes requested):
We can skip this rule since the feature gate is locked to true since k8s 1.25 and 1.25 is out of maintenance 

https://github.com/kubernetes/kubernetes/blob/88e994f6bf8fc88114c5b733e09afea339bea66d/pkg/features/kube_features.go#L1075

https://kubernetes.io/releases/

dimityrmirchev (approved):
/lgtm

PR #162, opened 2024-03-07
[DISA K8s STIG] Implement rule 242451 for all providers

AleksandarSavchev:
This PR implements DISA K8s STIGs rule 242451 for gardener, virtualgarden and managedk8s.

Part of #67

**Release note**:

```feature user
NONE
```


pkg/provider/virtualgarden/ruleset/disak8sstig/v1r11_ruleset.go
```diff
@@ -483,7 +483,7 @@ func (r *Ruleset) registerV1R11Rules(ruleOptions map[string]config.RuleOptionsCo
 			noKubeletsMsg,
 			rule.Skipped,
 		),
-		&sharedv1r11.Rule242451{
+		&v1r11.Rule242451{
```
dimityrmirchev, line 486: Now that this rule is used only by the virtual garden provider shouldn't we just hardcode the selectors or at least use the correct defaults in the rule implementation?

dimityrmirchev (approved):
/lgtm

Issue #159, opened 2024-03-06
Diff between 2 diki reports 


JordanJordanov:
**What would you like to be added**:
Diff functionality (between 2 diki reports from the same type) which shall allow us to identify regressions and take corresponsing actions 

**Why is this needed**:
It is important to identify changes in compliance (e.g. newly fulfilled rules or rules which are not comliant any more)

PR #164, opened 2024-03-12
Add diki diff command to create difference json between reports

AleksandarSavchev:
This PR adds `diki diff` command, that creates a `json` containing the difference between 2 `json` outputs from `diki run`.

Part of #159 

**Release note**:

```feature user
A new command `diki report diff` was introduced, that creates a `json` containing the difference between two `json` outputs from `diki run`.
```

```breaking user
The functionality corresponding to the `diki report` command is now available under the `diki report generate` command.
```

```feature user
The commands `diki report generate` and `diki report diff` now accept a `--output` flag that can be used to specify a file where the output report should be written.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,15 @@ diki report output.json > report.hmtl
 diki report --distinct-by=gardener=id output1.json output2.json > report.hmtl
 ```
 
+#### Diff
+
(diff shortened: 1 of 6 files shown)


dimityrmirchev (changes requested):
Looks good overall. I have left some suggestions

dimityrmirchev (approved):
/lgtm

PR #165, opened 2024-03-13
[DISA K8s STIG] Remove node-files rule 

AleksandarSavchev:
Since all the rules that node-files checks are now implemented as stand alone rules (ref https://github.com/gardener/diki/issues/82) there is no need for the node-files rule.

Fixes #

**Release note**:

```feature user
NONE
```


dimityrmirchev (approved):
/lgtm

PR #212, opened 2024-04-12
[DISA K8s STIG] Fix rule 242452 for gardener provider

AleksandarSavchev:

Fixes #

**Release note**:

```bugfix user
A bug causing rule 242452 for gardener provider to check seed nodes instead of shoot nodes was fixed.
```


dimityrmirchev (approved):
/lgtm

PR #223, opened 2024-04-25
Add `args` field `opsPodLabels` to all providers

AleksandarSavchev:
This PR add the option to users to add their own labels to diki pods for the `gardener`, `managedk8s` and `virtual` provider. Diki pod specific labels cannot be overwritten. Users can add their labels by setting the `args` field `opsPodLabels` with their wanted labels.

Fixes #

**Release note**:

```feature user
A new field `.args.additionalOpsPodLabels` has been added to the `gardener`, `managedk8s` and `virtual` providers. The field contains key value pairs that will be added to the `diki` ops pods as additional labels.
```

### DIFF
--- a/example/config/managedk8s.yaml
+++ b/example/config/managedk8s.yaml
@@ -4,6 +4,8 @@ providers:                   # contains information about known providers
   metadata:
     foo: bar
   args:
+    # opsPodLabels: # pod labels that will be added to diki ops pods
+    #   foo: bar
     kubeconfigPath: /tmp/kubeconfig.config  # path to cluster admin kubeconfig
   rulesets:
   - id: disa-kubernetes-stig
--- a/example/config/managedk8s.yaml
+++ b/example/config/managedk8s.yaml
@@ -4,7 +4,7 @@ providers:                   # contains information about known providers
   metadata:
     foo: bar
   args:
-    # opsPodLabels: # pod labels that will be added to diki ops pods
+    # additionalOpsPodLabels: # pod labels that will be added to diki ops pods
     #   foo: bar
     kubeconfigPath: /tmp/kubeconfig.config  # path to cluster admin kubeconfig
   rulesets:
--- a/pkg/kubernetes/pod/pod.go
+++ b/pkg/kubernetes/pod/pod.go
@@ -48,8 +48,8 @@ type SimplePodExecutor struct {
 type SimplePodContext struct {
 	client client.Client
 	config *rest.Config
-	// PodLabels are labels to be added to the created pods
-	PodLabels map[string]string
+	// AdditionalPodLabels are labels to be added to the created pods. If the a label key is already set by the pod constructor function it is not overwritten.
+	AdditionalPodLabels map[string]string
 	// IntervalWait is the time between wait API calls.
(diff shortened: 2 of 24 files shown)


pkg/kubernetes/pod/pod.go
```diff
@@ -48,17 +48,20 @@ type SimplePodExecutor struct {
 type SimplePodContext struct {
 	client client.Client
 	config *rest.Config
+	// PodLabels are labels to be added to the created pods
+	PodLabels map[string]string
```
dimityrmirchev, line 52:
```suggestion
	// AdditionalPodLabels are labels to be added to the created pods. If the a label key is already set by the pod constructor function it is not overwritten.
	AdditionalPodLabels map[string]string
```

pkg/kubernetes/pod/pod.go
```diff
@@ -67,6 +70,11 @@ func NewSimplePodContext(client client.Client, config *rest.Config) (*SimplePodC
 // Create creates a Pod and waits for it to get in Running state.
 func (spc *SimplePodContext) Create(ctx context.Context, podConstructorFn func() *corev1.Pod) (PodExecutor, error) {
 	pod := podConstructorFn()
+	for label, value := range spc.PodLabels {
+		if _, ok := pod.Labels[label]; !ok {
+			pod.Labels[label] = value
```
dimityrmirchev, line 75: We should also check if Labels is nil and initialise it if it is.

example/config/managedk8s.yaml
```diff
@@ -4,6 +4,8 @@ providers:                   # contains information about known providers
   metadata:
     foo: bar
   args:
+    # opsPodLabels: # pod labels that will be added to diki ops pods
```
dimityrmirchev, line 7: I think that we should rename this so that it is clear that these labels are added to an existing set of labels.

```suggestion
    # additionalOpsPodLabels: # pod labels that will be added to diki ops pods
```

dimityrmirchev (changes requested):
Just a few small things otherwise looks good

pkg/kubernetes/pod/pod.go
```diff
@@ -48,25 +48,37 @@ type SimplePodExecutor struct {
...
+	AdditionalPodLabels map[string]string
 	// IntervalWait is the time between wait API calls.
 	IntervalWait time.Duration
 	// TimeoutWait is the time waited for a pod to reach Running state or be deleted.
 	TimeoutWait time.Duration
 }
 
 // NewSimplePodContext creates a new SimplePodContext.
-func NewSimplePodContext(client client.Client, config *rest.Config) (*SimplePodContext, error) {
+func NewSimplePodContext(client client.Client, config *rest.Config, podLabels map[string]string) (*SimplePodContext, error) {
```
dimityrmirchev, line 60:
```suggestion
func NewSimplePodContext(client client.Client, config *rest.Config, additionalPodLabels map[string]string) (*SimplePodContext, error) {
```

dimityrmirchev (changes requested):

dimityrmirchev (approved):
/lgtm

PR #233, opened 2024-05-08
Add diki release version to generated json report

AleksandarSavchev:
This PR adds additional metadata to generated json Diki report. The new metadata is `dikiVersion`, which contains the release version of Diki.

Fixes #

**Release note**:

```feature user
The generated `json` report summary now contains a `dikiVersion` containing the release version of `Diki`.
```

### DIFF
--- a/pkg/report/report.go
+++ b/pkg/report/report.go
@@ -12,6 +12,8 @@ import (
 	"slices"
 	"time"
 
+	"k8s.io/component-base/version"
+
 	"github.com/gardener/diki/pkg/provider"
 	"github.com/gardener/diki/pkg/rule"
 	"github.com/gardener/diki/pkg/ruleset"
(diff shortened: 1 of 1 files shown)


dimityrmirchev (commented):
Shouldn't we also display this in the html reports 🤔 ?

AleksandarSavchev:
> Shouldn't we also display this in the html reports 🤔 ?

If we want to add `dikiVersion` to html reports we need to discuss if we want the html reports to contain the json reports' `dikiVersion` or the version of the code that generated the html report.
Also, if we choose to use the json's `dikiVersion` should we disallow merging of json reports of different diki versions or which diki version will we present in that case? Currently we can merge diki reports regardless of their `dikiVersion`.

dimityrmirchev:
Fair enough. We can discuss this at a later stage.

dimityrmirchev (approved):
/lgtm

PR #235, opened 2024-05-09
Add report metadata

AleksandarSavchev:
This PR creates a new `metadata` field in the Diki config file. The value of this field will be present in the json summary reports as report metadata. This metadata can be viewed in the html report as well, in merged html reports the metadata is only previewed if all merged reports have identical metadata.

Fixes #

**Release note**:

```feature user
Report metadata can now be added by setting the `metadata` field in the Diki config file.
```


cmd/diki/app/app.go
```diff
@@ -339,6 +339,9 @@ func runCmd(ctx context.Context, providerCreateFuncs map[string]provider.Provide
 			if dikiConfig.Output != nil && len(dikiConfig.Output.MinStatus) > 0 {
 				reportOpts = append(reportOpts, report.MinStatus(dikiConfig.Output.MinStatus))
 			}
+			if dikiConfig.Metadata != nil && len(dikiConfig.Metadata) > 0 {
```
dimityrmirchev, line 342:
```suggestion
			if len(dikiConfig.Metadata) > 0 {
```

cmd/diki/app/app.go
```diff
@@ -364,6 +367,9 @@ func runCmd(ctx context.Context, providerCreateFuncs map[string]provider.Provide
...
```
dimityrmirchev, line 370:
```suggestion
			if len(dikiConfig.Metadata) > 0 {
```

cmd/diki/app/app.go
```diff
@@ -387,6 +393,9 @@ func runCmd(ctx context.Context, providerCreateFuncs map[string]provider.Provide
...
```
dimityrmirchev, line 396:
```suggestion
			if len(dikiConfig.Metadata) > 0 {
```

pkg/report/render.go
```diff
@@ -46,12 +49,20 @@ func NewHTMLRenderer() (*HTMLRenderer, error) {
 		_, ok := m[k]
 		return ok
 	}
+	yamlFormat := func(m map[string]any) string {
+		yaml, err := yaml.Marshal(m)
+		if err != nil {
+			log.Println(err)
```
dimityrmirchev, line 55:
```suggestion
			return err.String()
```

Let's show the error in the report if there is such

pkg/report/report.go
```diff
@@ -79,6 +81,14 @@ func (ms MinStatus) ApplyToReport(opts *ReportOptions) {
 	}
 }
 
+// Metadata is additional report values.
+type Metadata map[string]any
+
+// ApplyToReport implements ReportOption.
+func (md Metadata) ApplyToReport(opts *ReportOptions) {
+	opts.Metadata = md
```
dimityrmirchev, line 89: Can't we use maps.Clone here?

dimityrmirchev (changes requested):
A few minor comments, other looks good. Can you please attach an image of how the new changes look in the html report?

AleksandarSavchev:
> Can you please attach an image of how the new changes look in the html report?

Yes, here are two examples:
image
image

dimityrmirchev (approved):
/lgtm

PR #249, opened 2024-06-11
Add `format` flag to `report generate` command

AleksandarSavchev:

Fixes #

**Release note**:

```feature user
The `report generate` command now accepts a `--format` flag which determines the output format of the generated report. It can be set to one of `html` or `json`. Defaults to `html`.
```

### DIFF
--- a/cmd/diki/app/app.go
+++ b/cmd/diki/app/app.go
@@ -143,7 +143,7 @@ func addRunFlags(cmd *cobra.Command, opts *runOptions) {
 
 func addReportGenerateFlags(cmd *cobra.Command, opts *generateOptions) {
 	cmd.PersistentFlags().Var(cliflag.NewMapStringString(&opts.distinctBy), "distinct-by", "If set generates a merged report. The keys are the IDs for the providers which the merged report will include and the values are distinct metadata attributes to be used as IDs for the different reports.")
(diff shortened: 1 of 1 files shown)


cmd/diki/app/app.go
```diff
@@ -143,6 +143,7 @@ func addRunFlags(cmd *cobra.Command, opts *runOptions) {
 
 func addReportGenerateFlags(cmd *cobra.Command, opts *generateOptions) {
 	cmd.PersistentFlags().Var(cliflag.NewMapStringString(&opts.distinctBy), "distinct-by", "If set generates a merged report. The keys are the IDs for the providers which the merged report will include and the values are distinct metadata attributes to be used as IDs for the different reports.")
+	cmd.PersistentFlags().StringVar(&opts.format, "format", "html", "Format for the merged report, dependent on 'distinct-by' flag to be set to take effect. Format can be 'html' or 'json'. Defaults to 'html'.")
```
dimityrmirchev, line 146: Cant we always have this flag taking an effect? If value is set to json without "--distinct-by" then we just write the report in `json` format. It does not make much sense to perform this operation, but it consistent. I do not like having the two flags that tightly connected.

AleksandarSavchev: For consistency we can add this option where when `--format=json` is set, but `--distinct-by` is not set we simply output the same json report as the input. I will now add it.

dimityrmirchev (approved):
/lgtm

PR #252, opened 2024-06-14
Fix bug causing rules to Error with "could not find files in foo"

AleksandarSavchev:
This PR fixes a bug causing rules to Error with "could not find files in foo" when there were no regular files. This bug was caused due to using the `ls` command to search for any not checked files. The `ls` command also returns special files and directories which caused the function to error when there were no regular files.

Fixes #

**Release note**:

```feature user
A bug causing rules that check files on nodes to error with `could not find files in foo` when there were no regular files in `foo` was fixed.
```


dimityrmirchev (approved):
/lgtm

PR #255, opened 2024-06-20
Return Errored check result when kube-proxy pod cannot be found

AleksandarSavchev:
This PR improved `kube-proxy` rules by making them return `Errored` check result when the `kube-proxy` pod cannot be found. `Errored` suits this case better since the rule could not be checked, hence it cannot be `Failed`.

Fixes #

**Release note**:

```other user
Rules that cannot find `kube-proxy` pod now return `Errored` check result.
```


dimityrmirchev (approved):
/lgtm

PR #257, opened 2024-06-24
Retry rules that Error with known messages for gardener provider

AleksandarSavchev:
**What this PR does / why we need it**:
This PR adds a new struct `RetryableRule` which is used to wrap a `Rule` and retry it's `Run()` method if a `retryCondition` is met.
A `retryCondition` func can be created by regexes using `RetryConditionFromRegex()`. This PR also adds regexes for 3 know transient errors in `pkg/rule/retryablerule/retryable_errors.go`
A new `args` field is introduced for `rulesets`, where ruleset specific arguments can be set. For `gardener` provider's `disa-kubernetes-stig` the `maxRetries` settings has been added

**Which issue(s) this PR fixes**:
Part of #253
Such changes should also be added to the `virtualgarden` and `managedk8s` providers.

**Special notes for your reviewer**:

**Release note**:

```feature user
A new `args` field is introduced for `rulesets`, where ruleset specific arguments can be set.
```
```feature user
For `gardener` provider's `disa-kubernetes-stig` ruleset, rule execution can now be retried if their `checkResults` contains a known `Errored` message.
```
```feature user
For `gardener` provider's `disa-kubernetes-stig` ruleset, the `args` field has been enhanced with `maxRetries` setting, which sets the number of maximum retries for rule runs. Defaults to 1.
```

PR #259, opened 2024-06-26
Retry rules that Error with known messages for managedk8s and virtualgarden providers

### LINKED ISSUE #253: Retry rules that Error with known messages
Rules can return Errored checkResults with with messages that suggest the error to be transient. We should retry the rule execution of rules that have errored with known to be transient error messages.

- [x] Implement `RetryableRule` wrapper struct #257 
- [x] Use `RetryableRule` for `gardener` provider #257 
- [x] Use `RetryableRule` for `virtualgarden` provider #259
- [x] Use `RetryableRule` for `managedk8s` provider #259

Command executions may error with `timeout` or `connection reset`, such errors should be reset by the `Execute()` method of `PodExecutor`

- [x] Improve `PodExecutor`'s `Execute()`  to retry timeouts and `connection reset` errors #260

**Why is this needed**:
This would help avoid transient errors.

AleksandarSavchev:
This PR adds retry logic to rules in the `managedk8s` and `virtualgarden` providers `disa-kubernetes-stig` ruleset. It adds them in a similar way to #257 

Part of #253 

**Release note**:

```feature user
The `disa-kubernetes-stig` ruleset rules for `virtualgarden` and `managedk8s` providers can now be retried if their results contain a known `Errored` message.
```
```feature user
The `disa-kubernetes-stig` ruleset config for `virtualgarden` and `managedk8s` providers has been enhanced with `maxRetries` setting, which sets the number of maximum retries for rule runs. Defaults to 1.
```

### DIFF
--- a/example/config/managedk8s.yaml
+++ b/example/config/managedk8s.yaml
@@ -11,6 +11,8 @@ providers:                   # contains information about known providers
   - id: disa-kubernetes-stig
     name: DISA Kubernetes Security Technical Implementation Guide
     version: v1r11
+    # args:
+    #   maxRetries: 1 # number of maximum rule run retries. Defaults to 1 
     ruleOptions:
     # - ruleID: "242376"
     #   skip:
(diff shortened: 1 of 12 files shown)


dimityrmirchev:
Can we combine the release notes so we have 2 instead of 4?

AleksandarSavchev:
> Can we combine the release notes so we have 2 instead of 4?

Yes, just edited the notes

dimityrmirchev (approved):
/lgtm

PR #261, opened 2024-06-28
Return Errored check result when pod cannot be found

AleksandarSavchev:
`Errored` suits this case better since the rule could not be checked, hence it cannot be `Failed`.

Fixes #

**Release note**:

```other user
Rules that cannot find specific pod now return `Errored` check result.
```


dimityrmirchev (approved):
/lgtm

PR #264, opened 2024-07-01
[DISA K8s STIG] Add kubeProxyDisabled option to rules

AleksandarSavchev:
This PR add a new rule option `kubeProxyDisabled` to rules that check the `kube-proxy` while checking other components as well. When the `kubeProxyDisabled` is set to `true` the `kube-proxy` check will be skipped.

This change is needed since a cluster can work without `kube-proxy` and adding the option will provide a way for rules that check not only the `kube-proxy` to skip only the `kube-proxy` check.

Fixes #

**Release note**:

```feature user
A new rule option `kubeProxyDisabled` is added to rules that check many components, which contain the `kube-proxy`. Setting this option to `true` would skip only the `kube-proxy` check in the rule. Defaults to `true`.
```

### DIFF
--- a/pkg/provider/gardener/ruleset/disak8sstig/v1r11/242400.go
+++ b/pkg/provider/gardener/ruleset/disak8sstig/v1r11/242400.go
@@ -38,7 +38,12 @@ type Rule242400 struct {
 	ClusterV1RESTClient   rest.Interface
 	ClusterPodContext     pod.PodContext
 	ControlPlaneNamespace string
+	Options               *Options242400
 	Logger                provider.Logger
+}
+
+type Options242400 struct {
+	KubeProxyDisabled bool `json:"kubeProxyDisabled" yaml:"kubeProxyDisabled"`
 }
 
 func (r *Rule242400) ID() string {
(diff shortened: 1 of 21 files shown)


AleksandarSavchev:
> Should we also include this option for other provider rules?

This change will be added to the `managedk8s` provider as well. For the `virtualgarden` one we do not check `kube-proxy`.

pkg/provider/gardener/ruleset/disak8sstig/v1r11/242400.go
```diff
@@ -105,7 +107,42 @@ func (r *Rule242400) Run(ctx context.Context) (rule.RuleResult, error) {
...
+		case allAlpha:
+			checkResults = append(checkResults, rule.FailedCheckResult(fmt.Sprintf("Option %s set to not allowed value.", option), target))
+		default:
+			checkResults = append(checkResults, rule.PassedCheckResult(fmt.Sprintf("Option %s set to allowed value.", option), target))
+		}
+	}
+
 	// kube-proxy check
+	if r.Options != nil && r.Options.KubeProxyDisabled {
+		checkResults = append(checkResults, rule.AcceptedCheckResult("Kube-proxy is disabled for cluster.", rule.NewTarget()))
```
dimityrmirchev, line 138:
```suggestion
		checkResults = append(checkResults, rule.AcceptedCheckResult("kube-proxy check is skipped.", rule.NewTarget()))
```

dimityrmirchev (approved):
/lgtm

PR #265, opened 2024-07-01
[DISA K8s STIG] Rule 242442 for gardener provider check image per namespace

AleksandarSavchev:
This PR improves `disa-kubernetes-stig` ruleset rule `242442` for `gardener` provider by making it check Pod's images per namespace.

The change brings the rules closer to the actual rule description of the stig found [here](https://www.stigviewer.com/stig/kubernetes/2023-08-29/finding/V-242442). In there rule it is wanted for old components to be removed when their updated versions have been installed. It is reasonable to assume that if a component is being updated it will happen the updated version will be in the same namespace as the old component

Fixes #

**Release note**:

```other user
Improved `disa-kubernetes-stig` ruleset rule `242442` for `gardener` provider to check pod images per namespace.
```

### DIFF
--- a/pkg/provider/gardener/ruleset/disak8sstig/v1r11/242442.go
+++ b/pkg/provider/gardener/ruleset/disak8sstig/v1r11/242442.go
@@ -37,14 +37,12 @@ func (r *Rule242442) Name() string {
 }
 
 func (r *Rule242442) Run(ctx context.Context) (rule.RuleResult, error) {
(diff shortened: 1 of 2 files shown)


dimityrmirchev (commented):
Can you please justify this change and why we are doing it in the PR description? Other looks good.

/lgtm

dimityrmirchev (approved):

PR #287, opened 2024-08-21
[DISA K8s STIG] Add support for version v2r1

AleksandarSavchev:
This PR adds support for the latest version of DISA K8s STIG `v2r1`. To avoid duplication of rule implementations, the rules that have not been changed between versions are now shared per provider and moved to the `pkg/provider/*/ruleset/disak8sstig/rules` dir.
Version `v2r1` is a special case where there are no new rules or rule changes between `v2r1` and the previous version `v1r11`. As such the `pkg/provider/*/ruleset/disak8sstig/v1r11` directories are completely removed. For future DISA K8s STIG releases the rule implementations of the latest version will be kept in the `pkg/provider/*/ruleset/disak8sstig/rules` dir, and there will be specific directories for previous versions with rule differences.

Fixes #

**Release note**:

```feature user
Diki now supports DISA Kubernetes STIG version `v2r1`
```


dimityrmirchev (approved):
/lgtm

PR #318, opened 2024-10-14
Generated summary reports should contain provider metadata

### LINKED ISSUE #296: Generated summary reports should contain provider metadata
**What happened**:
Running diki with `--ruleset-id` and `--ruleset-version` flags set generated a `json` report that does not contain the provider's metadata field

It is expected the generated `json` report to contain the provaider's `metadata` field.

To reproduce the bug simply run `diki` with the `--ruleset-id` and `--ruleset-version` flags set. The diki config should contains at least one provider with the `metadata` field.
The generated summary report is created without the `metadata` field for the `provider`.

georgibaltiev:
This PR updates the constructor of the ProviderResult struct when initialized (in the case we have both ruleset-id and ruleset-version flags specified) by adding the missing metadata field from the provider in question that is used for the ruleset run.

Fixes #296 

NONE

**Release note**:

```bugfix user
A bug causing generated summary reports to not contain `provider.metadata`, when `--ruleset-id` and `--ruleset-version` flags are set, was fixed.
```

### DIFF
--- a/cmd/diki/app/app.go
+++ b/cmd/diki/app/app.go
@@ -404,7 +404,7 @@ func runCmd(ctx context.Context, providerCreateFuncs map[string]provider.Provide
 		if err != nil {
 			return err
 		}
-		providerResults := []provider.ProviderResult{{ProviderID: p.ID(), ProviderName: p.Name(), RulesetResults: []ruleset.RulesetResult{res}}}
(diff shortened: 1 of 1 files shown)


AleksandarSavchev (approved):
Thanks!
/lgtm

PR #325, opened 2024-10-22
Check for proxy container if kube-proxy is not present

georgibaltiev:
**What this PR does / why we need it**:
This PR modifies the GetContainerCommand and GetContainerID util functions to be variadic and accept more than one possible container name for matching. Along with those modifications, rules that utilize the functions (242400, 242447, 242448) are also modified to search for "proxy" containers if "kube-proxy" containers are absent. Additional unit tests are added to validate the new modification.

**Which issue(s) this PR fixes**:
Fixes #314 

**Special notes for your reviewer**:
NONE

**Release note**:

```bugfix user
A bug causing DISA Kubernetes STIG rules that check the `kube-proxy` container to fail to find the container, when the container name is `proxy`, was fixed.
```

PR #326, opened 2024-10-23
Add additional comment suggestions to the GetContainer util methods

### LINKED ISSUE #314: Check for `proxy` container if `kube-proxy` is not present
**What happened**:
Rules that check the `kube-proxy` pods depend on the pods having a `kube-proxy` container. Ex.:
https://github.com/gardener/diki/blob/1529d1bbbc1a195da4adac46561b427c5a65145d/pkg/provider/managedk8s/ruleset/disak8sstig/rules/242400.go#L205

I would like that in all instances where we search for the `kube-proxy` container to check for `proxy` container if the former is not present

**Anything else we need to know**:

**Environment**:

georgibaltiev:
This PR adds additional comment suggestions that remained unaddressed from PR #325 

**Release note**:

```
NONE
```

### DIFF
--- a/pkg/internal/utils/utils.go
+++ b/pkg/internal/utils/utils.go
@@ -163,8 +163,8 @@ func GetMountedFilesStats(
 	return stats, err
 }
 
-// GetContainerID returns the container ID specified in the container status by container name
-// The first container name that exists in status is selected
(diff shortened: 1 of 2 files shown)


dimityrmirchev (changes requested):
One small change request. The container command is not retrieved from the pod container status, but rather from the containers list

PR #338, opened 2024-11-01
Refactor label matching for ETCD pods

### LINKED ISSUE #337: Rules that examine the ETCD pods do not select them correctly
**What happened**:
Rules that look for ETCD pods in order to be evaluated return an Errored check result and a message that the pods cannot be found (in an environment where ETCD Stateful Sets are present for example)

The aforementioned rules should not result in an error and should be able to select the ETCD pods correctly.

Run any of the DISA rules that examine the ETCD pods (e.g. 242459) with the gardener provider

**Anything else we need to know**:

**Environment**:

georgibaltiev:
This PR adds a second for-loop check to find ETCD pods by matching them with their new labels, amended by the ETCD druid operator's version 0.23.0 ([reference](https://github.com/gardener/etcd-druid/pull/777)).
The second loop is done only if no ETCD pods were found with the old labels. 
The errored check results can still display the old labels if old ETCD pods are found only partially, but if no ETCD pods are found the default behavior will be for the errored checkResults to specify the new labels.

Fixes #337  

**Release note**:

```bugfix user

```


pkg/shared/ruleset/disak8sstig/rules/242445.go
```diff
@@ -28,14 +28,16 @@ import (
...
-	Options            *option.FileOwnerOptions
-	Logger             provider.Logger
+	InstanceID                   string
+	Client                       client.Client
+	Namespace                    string
+	PodContext                   pod.PodContext
+	ETCDMainSelector             labels.Selector
+	ETCDMainNewVersionSelector   labels.Selector
+	ETCDEventsSelector           labels.Selector
+	ETCDEventsNewVersionSelector labels.Selector
```
AleksandarSavchev, line 38:
```suggestion
```

This rule is shared between `gardener` and `virtualgarden` provider, there is no need for these selectors here. WDYT @dimityrmirchev @georgibaltiev?

If you agree, we should also remove them for shared rule `242459`

pkg/shared/ruleset/disak8sstig/rules/242459.go
```diff
@@ -47,7 +49,10 @@ func (r *Rule242459) Name() string {
 func (r *Rule242459) Run(ctx context.Context) (rule.RuleResult, error) {
 	checkResults := []rule.CheckResult{}
 	etcdMainSelector := labels.SelectorFromSet(labels.Set{"instance": "etcd-main"})
+	etcdMainNewVersionSelector := labels.SelectorFromSet(labels.Set{"app.kubernetes.io/part-of": "etcd-main"})
+
```
AleksandarSavchev, line 53:
```suggestion
```

pkg/provider/gardener/ruleset/disak8sstig/rules/242451.go
```diff
@@ -98,26 +100,44 @@ func (r *Rule242451) Run(ctx context.Context) (rule.RuleResult, error) {
...
 			for _, p := range allSeedPods {
 				if podSelector.Matches(labels.Set(p.Labels)) && p.Namespace == r.ControlPlaneNamespace {
 					pods = append(pods, p)
 				}
 			}
 
 			if len(pods) == 0 {
-				checkResults = append(checkResults, rule.ErroredCheckResult("pods not found", seedTarget.With("namespace", r.ControlPlaneNamespace, "selector", podSelector.String())))
 				continue
 			}
```
AleksandarSavchev, line 118: We should not remove the check results here. It is possible for 0 pods to be found in matching the `etcd-main` selector, but then to be pods matching the `etcd-events` selector. In that case we lose Errored Check results.

Please keep the the checkResults from here in a new slice and then add them after the loop if the `len()` of `checkPods` is `> 0`. There should also be a unit test for this. Do this for other rules as well.

AleksandarSavchev (changes requested):
Please add unit tests, where the `instance` label is used.

pkg/provider/gardener/ruleset/disak8sstig/rules/242400.go
```diff
@@ -147,7 +147,7 @@ func (r *Rule242400) Run(ctx context.Context) (rule.RuleResult, error) {
 	if err != nil {
 		checkResults = append(checkResults, rule.ErroredCheckResult(err.Error(), shootTarget.With("kind", "podList")))
 	} else {
-		var pods []corev1.Pod
+		pods := []corev1.Pod{}
```
AleksandarSavchev, line 150:
```suggestion
		var pods []corev1.Pod
```
Please stick to the following convention as requested in: https://github.com/gardener/diki/pull/338#discussion_r1827385464

pkg/provider/gardener/ruleset/disak8sstig/rules/242451.go
```diff
@@ -98,26 +104,49 @@ func (r *Rule242451) Run(ctx context.Context) (rule.RuleResult, error) {
...
-			checkPods    []corev1.Pod
-			podSelectors = []labels.Selector{etcdMainSelector, etcdEventsSelector}
+			checkPods       []corev1.Pod
+			podOldSelectors = []labels.Selector{etcdMainOldSelector, etcdEventsOldSelector}
+			podSelectors    = []labels.Selector{etcdMainSelector, etcdEventsSelector}
 		)
 
-		for _, podSelector := range podSelectors {
-			var pods []corev1.Pod
+		oldSelectorCheckResults := []rule.CheckResult{}
```
AleksandarSavchev, line 112: Please move it to the `var` block.

pkg/provider/gardener/ruleset/disak8sstig/rules/242451_test.go
```diff
@@ -485,4 +564,5 @@ tlsCertFile: /var/lib/certs/tls.crt`
 				rule.ErroredCheckResult("could not retrieve kubelet config: bar", rule.NewTarget("cluster", "shoot", "name", "diki-242451-bbbbbbbbbb", "namespace", "kube-system", "kind", "pod")),
 			}),
 	)
+
```
AleksandarSavchev, line 567:
```suggestion
```

pkg/shared/ruleset/disak8sstig/rules/242445.go
```diff
@@ -28,14 +28,12 @@ import (
 var _ rule.Rule = &Rule242445{}
 
 type Rule242445 struct {
-	InstanceID         string
-	Client             client.Client
-	Namespace          string
-	PodContext         pod.PodContext
-	ETCDMainSelector   labels.Selector
-	ETCDEventsSelector labels.Selector
-	Options            *option.FileOwnerOptions
```
AleksandarSavchev, line 37: After a joint review with @dimityrmirchev we found that it should be reverted to the old implementation from https://github.com/gardener/diki/pull/338#discussion_r1827411256

This would handle the etcd pods in the virtual cluster since their label values contain the prefix `virtual-garden-` and this should be able to be configured when registering the rules for the virtualgarden provider.

We should return them for both shared rules.

AleksandarSavchev (changes requested):
Please add the suggested changes to all instances.

AleksandarSavchev (commented):
/lgtm

AleksandarSavchev (approved):

PR #344, opened 2024-11-05
Do not report empty targets

dimityrmirchev:
Empty targets can be seen in reports.

```json
{
  "targets": [
    {}
  ]
}
```

With this change this should no longer be the case.

Fixes #

**Release note**:

```bugfix user
A bug that was causing reports to sometimes include empty targets was fixed.
```


AleksandarSavchev (approved):
/lgtm

PR #346, opened 2024-11-06
Use nil slices instead of empty initialization

dimityrmirchev:
Code improvements. See https://github.com/uber-go/guide/blob/master/style.md#nil-is-a-valid-slice

Fixes #

**Release note**:

```feature user

```


AleksandarSavchev (approved):
/lgtm

PR #350, opened 2024-11-07
Reduce entered namespaces by `nsenter`

AleksandarSavchev:
This PR reduces the entered namespaces by `nsenter` to only the required namespace for `diki-ops` pod. This is done to avoid error resulted from not having permissions to enter a specific namespace.

It has been validated that diki works with only the mount `-m` namespace, if other namespaces are required in the future we can add them when they are needed.

Fixes #

**Release note**:

```feature user
NONE
```


dimityrmirchev (approved):
Thank you!

/lgtm

PR #351, opened 2024-11-07
Improve DISA K8s STIG rule 242382 to also check `AuthorizationConfiguration`

AleksandarSavchev:
With the introduction of [StructuredAuthorization](https://kubernetes.io/docs/reference/access-authn-authz/authorization/#configuring-the-api-server-using-an-authorization-config-file) when `authorization-mode` cannot be set when `authorization-config` is set. This PR improves DISA K8s STIG rule 242382 by also checking the specified `AuthoricationConfiguration` file.

Fixes #

**Release note**:

```feature user
DISA K8s STIG rule 242382 now also checks `AuthorizationConfiguration` when it is set via `authorization-config`.
```

### DIFF
--- a/pkg/internal/utils/set.go
+++ b/pkg/internal/utils/set.go
@@ -27,6 +27,20 @@ func Subset(s1, s2 []string) bool {
 	return true
 }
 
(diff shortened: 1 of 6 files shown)


dimityrmirchev (approved):
/lgtm

PR #354, opened 2024-11-14
Add Severity to All Implemented Rules

### LINKED ISSUE #348: Add Severity to Rules
Currently for implemented rules we have the rule's severity as part of the rule's Name. Ex:
https://github.com/gardener/diki/blob/b5441c36985d525ff08dfec38fc8fe145d174452/pkg/shared/ruleset/disak8sstig/rules/242386.go#L31

I would like the `Rule` interface to be extended with a `Severity` method which returns the rule's severity and to remove the severity from the output of the `Name()` method.

https://github.com/gardener/diki/blob/b5441c36985d525ff08dfec38fc8fe145d174452/pkg/rule/rule.go#L13-L18

georgibaltiev:
This PR adds a Severity field to all Rules that had it specified in their Name.

Fixes #348 

**Release note**:

```feature user
Severity level has been set to all current rules.
```


dimityrmirchev (commented):
We can also check that all known rulesets register only rules that implement severity by checking 

```golang
if _, ok := rules[i].(rule.Severity); !ok {
			return fmt.Errorf("rule %s does not implement rule.Severity", rules[i].ID())
		}
```

dimityrmirchev (approved):
/lgtm

Issue #356, opened 2024-11-14
☂ Implement `Security Hardened Kubernetes Cluster` ruleset


AleksandarSavchev:
**What would you like to be added**:
`DISA Kubernetes Security Technical Implementation` Guide should be extended with a new ruleset `Security Hardened Kubernetes Cluster` which aims to additionally increase the security posture of a Kubernetes cluster. Thos ruleset should be added to the `managedk8s` provider.

- [X] Add `Security Hardened Kubernetes Cluster` ruleset guide https://github.com/gardener/diki/pull/342
- [x] Add `Security Hardened Kubernetes Cluster` ruleset to `managedk8s` provider https://github.com/gardener/diki/pull/359
- [x] Implement Rules
  - [x] 2000 Ingress and egress traffic must be restricted by default. https://github.com/gardener/diki/pull/383
  - [x] 2001 Containers must be forbidden to escalate privileges. https://github.com/gardener/diki/pull/375
  - [x] 2002 Storage Classes should have a "Delete" reclaim policy. https://github.com/gardener/diki/pull/387
  - [x] 2003 Pods should use only allowed volume types. https://github.com/gardener/diki/pull/391
  - [x] 2004 Limit the Services of type NodePort. https://github.com/gardener/diki/pull/376
  - [x] 2005 Container images must come from trusted repositories. https://github.com/gardener/diki/pull/380
  - [x] 2006 Limit the use of wildcards in RBAC resources. https://github.com/gardener/diki/pull/382
  - [x] 2007 Limit the use of wildcards in RBAC verbs. https://github.com/gardener/diki/pull/389
  - [X] 2008 Pods must not be allowed to mount host directories. https://github.com/gardener/diki/pull/371
- [x] Add new usage documentation for running the `Security Hardened Kubernetes Cluster` ruleset

Ruleset documentation: https://github.com/gardener/diki/blob/main/docs/rulesets/security-hardened-k8s/ruleset.md

PR #387, opened 2024-11-28
[Security Hardened Kubernetes Cluster] Rule 2002 implementation 

georgibaltiev:

Part #356 

**Release note**:

```feature user
Implementation for rule `2002` from the `security-hardened-k8s` ruleset for provider `managedk8s`.
```

### DIFF
--- a/pkg/provider/managedk8s/ruleset/securityhardenedk8s/rules/2002.go
+++ b/pkg/provider/managedk8s/ruleset/securityhardenedk8s/rules/2002.go
@@ -51,8 +51,6 @@ func (r *Rule2002) Run(ctx context.Context) (rule.RuleResult, error) {
 	for _, storageClass := range storageClasses {
 		target := rule.NewTarget("kind", "storageClass", "name", storageClass.Name)
 		switch {
-		case storageClass.ReclaimPolicy == nil:
-			checkResults = append(checkResults, rule.FailedCheckResult("StorageClass does not have a configured ReclaimPolicy.", target))
 		case storageClass.ReclaimPolicy != nil && *storageClass.ReclaimPolicy == corev1.PersistentVolumeReclaimDelete:
 			checkResults = append(checkResults, rule.PassedCheckResult("StorageClass has a Delete ReclaimPolicy set.", target))
 		default:
--- a/pkg/provider/managedk8s/ruleset/securityhardenedk8s/rules/2002.go
+++ b/pkg/provider/managedk8s/ruleset/securityhardenedk8s/rules/2002.go
@@ -1,0 +1,64 @@
+// SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package rules
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kubeutils "github.com/gardener/diki/pkg/kubernetes/utils"
+	"github.com/gardener/diki/pkg/rule"
+)
+
+var (
+	_ rule.Rule     = &Rule2002{}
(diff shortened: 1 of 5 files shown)


pkg/provider/managedk8s/ruleset/securityhardenedk8s/rules/2002_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/gardener/diki/pkg/rule"
+)
+
+var _ = Describe("#2002", func() {
+	var (
+		client            client.Client
+		plainStorageClass *storagev1.StorageClass
+
+		ctx = context.TODO()
+	)
```
AleksandarSavchev, line 29:
```suggestion
	var (
		client            client.Client
		plainStorageClass *storagev1.StorageClass
		
		deleteReclaimPolicy    corev1.PersistentVolumeReclaimPolicy = "Delete"
		notDeleteReclaimPolicy corev1.PersistentVolumeReclaimPolicy = "NotDelete"

		ctx = context.TODO()
	)
```
I suggest we use these hard-coded policies in the units tests instead of `corev1.PersistentVolumeReclaimDelete` and `corev1.PersistentVolumeReclaimRetain`. So that we can catch when there are changes in upstream.

AleksandarSavchev (changes requested):
Thanks! There are some small changes, otherwise looks good.

pkg/provider/managedk8s/ruleset/securityhardenedk8s/rules/2002.go
```diff
@@ -0,0 +1,63 @@
...
+	var checkResults []rule.CheckResult
+
+	for _, storageClass := range storageClasses {
+		switch {
+		case storageClass.ReclaimPolicy == nil:
+			checkResults = append(checkResults, rule.PassedCheckResult("StorageClass defaults to Delete ReclaimPolicy.", rule.NewTarget("kind", "storageClass", "name", storageClass.Name)))
+		case *storageClass.ReclaimPolicy == corev1.PersistentVolumeReclaimDelete:
+			checkResults = append(checkResults, rule.PassedCheckResult("StorageClass has a Delete ReclaimPolicy set.", rule.NewTarget("kind", "storageClass", "name", storageClass.Name)))
+		default:
+			checkResults = append(checkResults, rule.FailedCheckResult("StorageClass does not have a Delete ReclaimPolicy set.", rule.NewTarget("kind", "storageClass", "name", storageClass.Name)))
```
dimityrmirchev, line 58: You can also extract the target as variable at the beginning of the loop.
```suggestion
		case storageClass.ReclaimPolicy != nil && *storageClass.ReclaimPolicy == corev1.PersistentVolumeReclaimDelete:
			checkResults = append(checkResults, rule.PassedCheckResult("StorageClass has a Delete ReclaimPolicy set.", target))
		default:
			checkResults = append(checkResults, rule.FailedCheckResult("StorageClass does not have a Delete ReclaimPolicy set.", target))
```

dimityrmirchev (changes requested):
Just one request:

Storage classes are defaulted to have reclaim policy `Delete`. We should treat a `nil` reclaim policy as `Failed`.

pkg/provider/managedk8s/ruleset/securityhardenedk8s/rules/2002.go
```diff
@@ -0,0 +1,64 @@
...
+		return rule.Result(r, rule.PassedCheckResult("The cluster does not have any StorageClasses.", rule.NewTarget())), nil
+	}
+
+	var checkResults []rule.CheckResult
+
+	for _, storageClass := range storageClasses {
+		target := rule.NewTarget("kind", "storageClass", "name", storageClass.Name)
+		switch {
+		case storageClass.ReclaimPolicy == nil:
+			checkResults = append(checkResults, rule.FailedCheckResult("StorageClass does not have a configured ReclaimPolicy.", target))
```
dimityrmirchev, line 55: No need for this case. We cover it in the dafault path

dimityrmirchev (approved):
/lgtm

PR #397, opened 2024-12-06
Add Graceful Shutdown Implementation

### LINKED ISSUE #285: Diki should be able to terminate gracefully
The program should not immediately exit when it receives a `SIGTERM` signal. Ideally `diki` should be able to cleanup any temporary state that it can create in a target provider, i.e. when targeting kubernetes clusters temporary pods should be cleaned up before exit.

georgibaltiev:
This PR implements a graceful shutdown when Dikl receives a SIGINT or SIGTERM signal by refactoring the tool's process management. With the new changes, a cancellable context is used for the execution of the rule check, thereby adding an additional layer for managing the terminating signals. 
If the context is cancelled during the execution of a ruleset, it terminates all goroutines after their corresponding rule's execution ends, and returns an error value. Additionally, a similar mechanism has been added when the parent process iterates through a provider's registered rulesets.
The deferred pod deletion functions in the rules that utilize them, use an independent context, which triggers a 30 second timeout after the function is called. That way, the deletion will be uninterrupted by the SIGINT/SIGTERM signal.

Fixes #285 

**Release note**:

```feature user
Diki can now shutdown gracefully when interrupted.
```

### DIFF
--- a/cmd/diki/app/app.go
+++ b/cmd/diki/app/app.go
@@ -26,7 +26,7 @@ import (
 )
 
 // NewDikiCommand creates a new command that is used to start Diki.
-func NewDikiCommand(ctx context.Context, providerCreateFuncs map[string]provider.ProviderFromConfigFunc) *cobra.Command {
+func NewDikiCommand(providerCreateFuncs map[string]provider.ProviderFromConfigFunc) *cobra.Command {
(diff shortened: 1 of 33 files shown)


dimityrmirchev (approved):
/lgtm

PR #399, opened 2024-12-09
Fix `GetDeploymentPods` utils function when pod does not have `OwnerReference`

AleksandarSavchev:

Fixes #

**Release note**:

```bugfix user
A bug causing some rules to error when they encounter a `Pod` without an `OwnerReference` has been fixed.
```


dimityrmirchev (approved):
/lgtm

PR #406, opened 2024-12-13
Add a guide documentation for validating the Security Hardened Kubernetes ruleset against a Gardener Shoot Cluster

georgibaltiev:
**What this PR does / why we need it**:
This PR adds an example configuration for running a compliance check with the Security Hardened Kubernetes ruleset for the `managedk8s` provider against a Gardener Shoot Cluster, as well as an .MD containing a step-by-step breakdown of the process. 

**Which issue(s) this PR fixes**:
Part of #356

**Special notes for your reviewer**:

**Release note**:

```feature user

```

PR #409, opened 2024-12-17
Add a wildcard option for volumes contained in AcceptedPods for Rules 2003 and 2008

### LINKED ISSUE #407: Allow selecting all volume names via wildcard "*" in rule 2003 from Security Hardened K8s
Allow selecting all volume names via wildcard "*" in rule 2003 from Security Hardened K8s.

**Why is this needed**:
It would be easier to express "allow all gardener managed resources". See this PR as a reference https://github.com/gardener/diki/pull/406

georgibaltiev:
This PR contains a modification to the acceptedResources handling in Rules 2003 and 2008 of the Security Hardened Kubernetes Ruleset by adding the option to accept all volumes in an accepted pod by using a wildcard. 

Fixes #407 

**Release note**:

```feature user
Accepted volume names for rules 2003 and 2008 of the Security Hardened Kubernetes Ruleset can be generalized by using a wildcard in the configuration.
```

### DIFF
--- a/example/config/managedk8s.yaml
+++ b/example/config/managedk8s.yaml
@@ -197,6 +197,13 @@ providers:                   # contains information about known providers
     #       volumeNames:
     #       - "volume-a"
     #       - "volume-b"
+    #     - matchLabels:
+    #         foo: baz
+    #       namespaceMatchLabels:
+    #         foo: baz
+    #       justification: "justification"
+    #       volumeNames:
+    #       - "*" # a wildcard can be used to match against all volumes in an accepted pod
     # - ruleID: "2004"
     #   args:
     #     acceptedServices:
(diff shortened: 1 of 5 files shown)


dimityrmirchev (changes requested):
Please also add the proper example in the sample configuration file

georgibaltiev:
I will push the guide change to #406 to avoid merge conflicts later on.

dimityrmirchev:
I am not sure what you mean. I mean having the new wildcard value shown [here](https://github.com/gardener/diki/blob/1405cfc2b78ca17b2a80d803109e74e87fffbdce/example/config/managedk8s.yaml#L197)

On a similar note I think that we should also do the same changes for rule 2008. WDYT?

georgibaltiev:
I assumed that you meant the guide configuration that will be used by the stakeholders. I agree that 2008 could benefit from the same mechanism, so I will add the changes to that rule too.

PR #418, opened 2025-01-08
Add a minimal status option to the report generating command

### LINKED ISSUE #378: Add ability to set the minimum status of included checks in generated reports
I want to be able to set the minimum status of checks that are included in generated (html) reports. Not all users might be interested in all targets that are checked. Some users might want to only see violations/fails/errors. We already support a similar configuration option when running `diki` https://github.com/gardener/diki/blob/f09debf935132dc9a10e8150fa75dff89b3f6172/example/config/managedk8s.yaml#L213.

**Why is this needed**:
In order to limit the size of a generated html reports.

georgibaltiev:
This PR adds an additional flag to the diki report generate command that specifies the minimal status of the checks that should be included in the rendered report. This flag only discards check results that have a higher status when generating a regular or merged report.

Fixes #378 

**Release note**:

```feature user
`diki report generate` now has a --min-status flag that discards check results with a higher status than the argument.
```

### DIFF
--- a/cmd/diki/app/app.go
+++ b/cmd/diki/app/app.go
@@ -13,6 +13,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"slices"
 
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
@@ -22,6 +23,7 @@ import (
 	"github.com/gardener/diki/pkg/config"
 	"github.com/gardener/diki/pkg/provider"
 	"github.com/gardener/diki/pkg/report"
+	"github.com/gardener/diki/pkg/rule"
 	"github.com/gardener/diki/pkg/ruleset"
 )
 
@@ -144,6 +146,7 @@ func addRunFlags(cmd *cobra.Command, opts *runOptions) {
 func addReportGenerateFlags(cmd *cobra.Command, opts *generateOptions) {
 	cmd.PersistentFlags().Var(cliflag.NewMapStringString(&opts.distinctBy), "distinct-by", "If set generates a merged report. The keys are the IDs for the providers which the merged report will include and the values are distinct metadata attributes to be used as IDs for the different reports.")
 	cmd.PersistentFlags().StringVar(&opts.format, "format", "html", "Format for the output report. Format can be one of 'html' or 'json'.")
+	cmd.PersistentFlags().StringVar(&opts.minStatus, "min-status", "Passed", "If set specifies the minimal status that will be included in the generated report. Status can be one of 'Passed', 'Skipped', 'Accepted', 'Warning', 'Failed', 'Errored' or 'NotImplemented'")
 }
 
 func addReportDiffFlags(cmd *cobra.Command, opts *diffOptions) {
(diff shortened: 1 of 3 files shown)


pkg/report/report.go
```diff
@@ -240,3 +240,26 @@ func getChecks(checkResults []rule.CheckResult, opts *ReportOptions) []Check {
 	}
 	return checks
 }
+
+// DiscardCheckResultsBelowMinStatus removes all checks of the report, whose statuses are lower than the specified one.
+func (r *Report) DiscardCheckResultsBelowMinStatus(minStatus rule.Status) {
```
AleksandarSavchev, line 245:
```suggestion
// SetMinStatus sets minStatus of the report. It also removes all checks with status less than the specified one.
// If the new minStatus is less that the original one, nothing is changed.
func (r *Report) SetMinStatus(minStatus rule.Status) {
```

I think this would be a better name for the method. WDYT?

AleksandarSavchev (changes requested):
Thanks!

pkg/report/report.go
```diff
@@ -125,6 +125,26 @@ func (r *Report) WriteToFile(filePath string) error {
 	return os.WriteFile(filePath, data, 0600)
 }
 
+// SetMinStatus sets minStatus of the report. It also removes all checks with status less than the specified one.
+// If the new minStatus is less that the original one, nothing is changed.
+func (r *Report) SetMinStatus(minStatus rule.Status) {
```
dimityrmirchev, line 130: Can't we implement this as a report option?

cmd/diki/app/app.go
```diff
@@ -144,6 +146,7 @@ func addRunFlags(cmd *cobra.Command, opts *runOptions) {
 func addReportGenerateFlags(cmd *cobra.Command, opts *generateOptions) {
 	cmd.PersistentFlags().Var(cliflag.NewMapStringString(&opts.distinctBy), "distinct-by", "If set generates a merged report. The keys are the IDs for the providers which the merged report will include and the values are distinct metadata attributes to be used as IDs for the different reports.")
 	cmd.PersistentFlags().StringVar(&opts.format, "format", "html", "Format for the output report. Format can be one of 'html' or 'json'.")
+	cmd.PersistentFlags().StringVar(&opts.minStatus, "min-status", "Passed", "If set specifies the minimal status that will be included in the generated report. Status can be one of 'Passed', 'Skipped', 'Accepted', 'Warning', 'Failed', 'Errored' or 'NotImplemented'")
```
dimityrmirchev, line 149: We should mention the statuses in a sorted order stating which is the highest

dimityrmirchev (changes requested):

pkg/report/report.go
georgibaltiev: I will look into it.

pkg/report/report.go
```diff
@@ -125,6 +125,26 @@ func (r *Report) WriteToFile(filePath string) error {
 	return os.WriteFile(filePath, data, 0600)
 }
 
+// SetMinStatus sets minStatus of the report. It also removes all checks with status less than the specified one.
+// If the new minStatus is less that the original one, nothing is changed.
```
dimityrmirchev, line 129:
```suggestion
// If the new minStatus is less than the original one, nothing is changed.
```

pkg/report/report.go
dimityrmirchev: You can disregard my comment. I see that we already have this and you are making use of it.

dimityrmirchev (changes requested):

AleksandarSavchev (approved):
/lgtm

dimityrmirchev (approved):
/lgtm

PR #453, opened 2025-03-20
[DISA K8s STIG] Improve rule 242436

AleksandarSavchev:
This PR improves DISA STIG rule 242436 to accept default values of `enable-admission-plugins` as  `ValidatingAdmissionWebhook` is part of the defaults https://github.com/gardener/gardener/pull/11655

Fixes #

**Release note**:

```bugfix user
Fixed a bug which was causing DISA STIG rule 242436 to fail when `ValidatingAdmissionWebhook` is not set in `enable-admission-plugins` even though it is defaulted.
```


dimityrmirchev (approved):
/lgtm

PR #481, opened 2025-04-29
[DISA K8s STIG] Refactor rule 242459

AleksandarSavchev:
This PR changes the implementation of DISA K8s STIG rule 242459 to check etcd files for `644` permissions or more restrictive, as stated in [the rule](https://stigviewer.com/stigs/kubernetes/2024-08-22/finding/V-242459). Files in `/etcd/data` directory are expected to have a maximum of `600` permissions, as they should not be accessible by `other` group and require write permissions.

`600` permissions are the minimal required for files in `/etcd/data` since the `etcd` creates and manages these files.

`640` permissions are still being enforced for all `.key` & `.pem` files via diki's implementation of [rule 242467](https://stigviewer.com/stigs/kubernetes/2024-08-22/finding/V-242467).

`644` permissions should be good enough for the general case, files with such permissions will be readable by all users connected to the container, but most of the files will still be isolated on the host from users since the `kubelet` directory (which contains pod volume mounts) has `000` permissions [ref](https://github.com/gardener/gardener/blob/c6308703a6d84ca87cc038e1ddd35393c1aec35e/cmd/gardener-node-agent/app/bootstrappers/kubelet_bootstrap_kubeconfig.go#L61-L64)

Fixes #

**Release note**:

```feature user
DISA STIG rule 242459 now checks for `644` permissions or more restrictive, for files in `/etcd/data` directory it expects `600` permissions.
```

### DIFF
--- a/pkg/internal/utils/utils.go
+++ b/pkg/internal/utils/utils.go
@@ -29,6 +29,7 @@ type FileStats struct {
 	Permissions           string
 	UserOwner, GroupOwner string
 	FileType              string
+	Destination           string
 }
 
 // NewFileStats creates a new FileStats object from the result of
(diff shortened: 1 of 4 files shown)


shreyas-s-rao (approved):
Thanks @AleksandarSavchev 
/lgtm

dimityrmirchev (approved):
/lgtm

PR #496, opened 2025-05-20
Retry rules on not found object by `nerdctl`

AleksandarSavchev:
This PR adds retry on not found objects by `nerdctl`. In most cases these objects are the IDs of pod containers, which may may no longer exist when we try to find them with `nerdctl`

Fixes #

**Release note**:

```feature user
NONE
```


dimityrmirchev (approved):
/lgtm

PR #507, opened 2025-06-02
Use static values for `Target`s

### LINKED ISSUE #504: Use static values for `Target`s
I would like `CheckResult` to use static values for their `Target`s.

For example when Diki is checking a `Pod` it uses the `Pod`'s name  in the `Target`. If the `Pod` is part of a workload resource like a `Deployment` the pods might be recreated (if there are changes to the `Deployment`) or removed (downscaling), also new pods of the deployment may be created when upscaling. This problem can be solved if we use the ownerReference of the resource, which contains static values.

**Why is this needed**:
Improve consistency between diki runs.

AleksandarSavchev:

Fixes #504

**Release note**:

```feature user

```

### DIFF
--- a/pkg/provider/gardener/ruleset/disak8sstig/rules/242414.go
+++ b/pkg/provider/gardener/ruleset/disak8sstig/rules/242414.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"slices"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


pkg/provider/gardener/ruleset/disak8sstig/rules/242414.go
```diff
@@ -56,32 +57,45 @@ func (r *Rule242414) Run(ctx context.Context) (rule.RuleResult, error) {
...
 
 	shootPods, err := kubeutils.GetPods(ctx, r.ClusterClient, "", labels.NewSelector(), 300)
 	if err != nil {
 		return rule.Result(r, append(checkResults, rule.ErroredCheckResult(err.Error(), shootTarget.With("kind", "podList")))...), nil
 	}
+	filteredShootPods := kubeutils.FilterPodsByOwnerRef(shootPods)
+
+	shootReplicaSets, err := kubeutils.GetReplicaSets(ctx, r.ClusterClient, "", labels.NewSelector(), 300)
+	if err != nil {
+		return rule.Result(r, rule.ErroredCheckResult(err.Error(), rule.NewTarget("namespace", r.ControlPlaneNamespace, "kind", "repliaceSetList"))), nil
```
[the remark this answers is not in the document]
dimityrmirchev: I wonder if we should unify capitalization, i.e. write "Deployment" and not "deployment", "ReplicaSetList" and not "replicaSetList".

dimityrmirchev (commented):
I left some preliminary comments. Will do another review once these are addressed. I propose that you do not modify any more rules. These changes can be made in another PR.

pkg/provider/managedk8s/ruleset/disak8sstig/rules/242414.go
```diff
@@ -48,22 +49,28 @@ func (r *Rule242414) Run(ctx context.Context) (rule.RuleResult, error) {
 	if err != nil {
 		return rule.Result(r, rule.ErroredCheckResult(err.Error(), rule.NewTarget("kind", "podList"))), nil
 	}
+	filteredPods := kubeutils.FilterPodsByOwnerRef(pods)
+
+	replicaSets, err := kubeutils.GetReplicaSets(ctx, r.Client, "", labels.NewSelector(), 300)
+	if err != nil {
+		return rule.Result(r, rule.ErroredCheckResult(err.Error(), rule.NewTarget("kind", "repliaceSetList"))), nil
```
dimityrmirchev, line 56:
```suggestion
		return rule.Result(r, rule.ErroredCheckResult(err.Error(), rule.NewTarget("kind", "ReplicaSetList"))), nil
```

pkg/kubernetes/utils/utils_test.go
```diff
@@ -2342,4 +2415,119 @@ readOnlyPort: 222
...
+					{
+						APIVersion: "apps/v1",
+						Kind:       "ReplicaSet",
+						UID:        "1",
+					},
+				},
+			},
+			rule.NewTarget("kind", "Deployment", "name", "deployment", "namespace", "bar"),
+		),
+		Entry("should return correct target when objects has more than 1ownerReference",
```
dimityrmirchev, line 2509:
```suggestion
		Entry("should return correct target when objects has more than 1 ownerReference",
```

pkg/kubernetes/utils/utils_test.go
```diff
@@ -278,6 +278,79 @@ var _ = Describe("utils", func() {
 		})
 	})
 
+	Describe("#GFilterPodsByOwnerRef", func() {
```
dimityrmirchev, line 281:
```suggestion
	Describe("#FilterPodsByOwnerRef", func() {
```

dimityrmirchev (changes requested):
Just a couple of nits.

Can you please share a comparison of the targets before and after this change. No need to share the whole run output, but just an example.

PR #530, opened 2025-07-08
Refactor rules from the Managed Kubernetes provider to use stable targets

### LINKED ISSUE #504: Use static values for `Target`s

georgibaltiev:
This PR refactors rules, implemented specifically for the Virtual Garden provider to use static checkResult targets.

Additionally, other rules and util functions that have not been updated with the previous PRs are included in this scope.

Fixes #504

**Release note**:

```feature user
All Diki rules now return `checkResults` with stable values in `targets` instead of static ones.
```


AleksandarSavchev (approved):
/lgtm

PR #533, opened 2025-07-10
Remove old etcd labels

georgibaltiev:
Since the release of version `0.23.0` of the `etcd-druid` operator, label selectors on etcd-druid managed resources [no longer include](https://github.com/gardener/etcd-druid/pull/777) key values like `name` or `instance`.

Instead, more standard selectors like `app.kubernetes.io/managed-by` and `app.kubernetes.io/part-of` have been introduced.

Diki has used the old selectors to evaluate druid-managed resources, [but since this PR has been merged](https://github.com/gardener/diki/pull/338), both selector versions have been supported simultaneously.

This PR drops the support for the old `etcd-druid` selectors entirely.

NONE

**Release note**:

```feature user
Diki no longer supports evaluation for etcd-druid managed resources that are labeled with the instance and name selectors.
```


AleksandarSavchev (approved):
Thanks! It is always nice to get rid of some no longer needed technical dept.

PR #543, opened 2025-07-31
Configure rule options for rule 242442 of the DISA STIG ruleset

### LINKED ISSUE #539: Allow exceptions for certain images in Rule 242442 of the DISA STIG ruleset
Currently, Rule 242442 aims to report for Kubernetes components that have deployments with more than one version on a particular node. 
The evaluation is achieved by retrieving the images of all containers and comparing their versions/tags/digests.

This, albeit working properly, does not differentiate for example between a case where a singular pod has deployed multiple containers with differing image versions and separate deployments that are of different releases.

In order to be more specific, we should be able to exempt certain images from evaluation with a list of allowed images in the configuration file.

An example `ruleOption` configuration could look like the following:
```yaml
ruleOptions: 
- ruleID: "242442"
  allowedImages:
  - name: eu.gcr.io/image1
  - name: eu.gcr.io/image2
```

**Why is this needed**:
To improve stakeholder experience.

georgibaltiev:

This PR modifies rule 242442 of the DISA STIG to accept options that specify a list of images, that are expected to have multiple versions. If a container image appears as a finding, it will be listed as a `Warning` if it's name is listed in the rule options of the configuration file.

Users can configure their list of allowed images the following way:
```yaml
ruleOptions: 
- ruleID: "242442"
  expectedVersionedImages:
  - name: eu.gcr.io/image1
  - name: eu.gcr.io/image2
```

Fixes #539 

**Release note**:

```feature user
Users can now configure a list of expected images that could have multiple versions for Rule 242442 of the DISA STIG ruleset. Any image finding that is listed in the configuration will be described as a `Warning`.
```

### DIFF
--- a/pkg/provider/managedk8s/ruleset/disak8sstig/rules/242442.go
+++ b/pkg/provider/managedk8s/ruleset/disak8sstig/rules/242442.go
@@ -129,7 +129,7 @@ func (r *Rule242442) checkImages(pods []corev1.Pod, target rule.Target) []rule.C
 			if ref, ok := images[imageBase]; ok && ref != imageRef {
 				if _, reported := reportedImages[imageBase]; !reported {
 					reportedImages[imageBase] = struct{}{}
-					if r.Options != nil && slices.ContainsFunc(r.Options.ImageSelector.ExpectedVersionedImages, func(expectedImage option.ExpectedVersionedImage) bool {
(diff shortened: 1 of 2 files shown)


AleksandarSavchev (changes requested):
Thanks, just found some nits!

AleksandarSavchev (dismissed):

dimityrmirchev (approved):
/lgtm

PR #569, opened 2025-08-26
Use `ubuntu-latest` for all diki binaries

AleksandarSavchev:
The currently built diki binaries are broken for linux-arm systems:
``` bash
$ ./diki-linux-arm64
bash: ./diki-linux-arm64: cannot execute: required file not found
```

The runner version `ubuntu-24.04-arm` seems to be the cause.
This PR reverts the version to `ubuntu-latest`, which would build a working binary

Fixes #

**Release note**:

```feature user
NONE
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -34,13 +34,13 @@ jobs:
             runner: ubuntu-latest
           - os: linux
             arch: arm64
-            runner: ubuntu-24.04-arm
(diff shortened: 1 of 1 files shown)


dimityrmirchev:
Can you add a justification why do we need this?

AleksandarSavchev:
> Can you add a justification why do we need this?

Yes, just added some description

PR #580, opened 2025-09-04
Use `logr.FromSlogHandler` instead of slogr wrapper

AleksandarSavchev:
This PR removes the `slogr` wrapper struct in `cmd/internal/slogr` and replaces it with the use of `logr.FromSlogHandler`.

Fixes #

**Release note**:

```feature user
NONE
```


cmd/diki/app/app.go
```diff
@@ -407,10 +407,10 @@ func generateCmd(args []string, rootOpts reportOptions, opts generateOptions, lo
 	}
 }
 
-func runCmd(ctx context.Context, providerCreateFuncs map[string]provider.ProviderFromConfigFunc, opts runOptions, logger *slog.Logger) error {
+func runCmd(ctx context.Context, providerCreateFuncs map[string]provider.ProviderFromConfigFunc, opts runOptions, handler slog.Handler) error {
```
dimityrmirchev, line 410: Can't we pass the logger directly?
```suggestion
func runCmd(ctx context.Context, providerCreateFuncs map[string]provider.ProviderFromConfigFunc, opts runOptions, logger logr.Logger) error {
```

dimityrmirchev (changes requested):
One comment

PR #620, opened 2025-10-15
Enhance `diki run` command to run with a predefined configuration for the Managed Kubernetes provider

### LINKED ISSUE #556: Enhance `diki run` command to run with predefined configuration
I would like to simplify the `diki run` command by allowing it to be run without specifying a config file and using a predefined config on the k8s cluster targeted with the `KUBECONFIG` env var.

Since diki was designed to work on non-Kubernetes environments as well we can add a flag that specifies the environment (by default it would be set to `kubernetes`), e.g.:

```
diki run --env=kubernetes
```

**Why is this needed**:
Simplify first time user experience, increase adoption.

georgibaltiev:

/kind enhancement

This PR adds a default run configuration for the `Managed Kubernetes` provider. 
A `diki run` command, specified with the `--provider=managedk8s` flag will not require a run configuration file (`--config`), and if omitted, a default run configuration will be used for the scan.

This PR is needed to simplify the users's first-time interactions with Diki.

Fixes #556

**Release note**:

```feature user
The `diki run` command can be executed without specifying a run configuration when the provider is set to managedk8s.
```

### DIFF
--- a/cmd/diki/app/app.go
+++ b/cmd/diki/app/app.go
@@ -608,7 +608,7 @@ func getDikiConfig(opts runOptions, defaultConfigFuncs map[string]provider.Defau
 		return config.DikiConfig{}, fmt.Errorf("unknown provider: %s", opts.provider)
 	} else {
 		if defaultFunc == nil {
-			return config.DikiConfig{}, fmt.Errorf("provider %s cannot resolve to a default configuration. Configuration must be specified with the --config flag", opts.provider)
+			return config.DikiConfig{}, fmt.Errorf("provider %s cannot resolve to a default configuration, it must be specified with the --config flag", opts.provider)
 		}
 
 		dikiConfig := defaultFunc()


cmd/diki/app/app.go
```diff
@@ -589,3 +594,22 @@ func getProvidersFromConfig(c *config.DikiConfig, providerCreateFuncs map[string
...
+	} else {
+		if len(opts.provider) == 0 {
+			return nil, fmt.Errorf("--provider should be set when --config is omitted")
+		}
+
+		if defaultFunc, ok := defaultConfigFuncs[opts.provider]; !ok {
+			return nil, fmt.Errorf("unknown provider: %s", opts.provider)
+		} else {
+			if defaultFunc == nil {
+				return nil, fmt.Errorf("provider %s cannot resolve a default Diki configuration - it should be specified with the --config flag", opts.provider)
```
AleksandarSavchev:
```suggestion
				return nil, fmt.Errorf("provider %s cannot resolve to a default configuration. Configuration must be specified with the --config flag", opts.provider)
```

AleksandarSavchev (changes requested):
Some more nits

cmd/diki/app/app.go
```diff
@@ -589,3 +594,24 @@ func getProvidersFromConfig(c *config.DikiConfig, providerCreateFuncs map[string
...
+
+	if len(opts.provider) == 0 {
+		return config.DikiConfig{}, fmt.Errorf("--provider must be set when --config is omitted")
+	}
+
+	if defaultFunc, ok := defaultConfigFuncs[opts.provider]; !ok {
+		return config.DikiConfig{}, fmt.Errorf("unknown provider: %s", opts.provider)
+	} else {
+		if defaultFunc == nil {
+			return config.DikiConfig{}, fmt.Errorf("provider %s cannot resolve to a default configuration. Configuration must be specified with the --config flag", opts.provider)
```
dimityrmirchev:
```suggestion
			return config.DikiConfig{}, fmt.Errorf("provider %s cannot resolve to a default configuration, configuration must be specified with the --config flag", opts.provider)
```

dimityrmirchev (changes requested):
One nit

Issue #687, opened 2026-02-18
Add support for Merge of rule Options


AleksandarSavchev:
**What would you like to be added**:
Diki should be able to handle rule options configurations with multiple records of the same rule. This can be accomplished by extending the Options interface with a `Merge(opt Option` method.
https://github.com/gardener/diki/blob/9c26eee8c2c29db6ada1742c26c4cad6b43f67e5/pkg/shared/kubernetes/option/options.go#L20-L24

Ideally there should be a way to configure a merge strategy. There should be 2 such strategies: `replace` (override) and `combine`
- `replace` would only use the last defined rule option for the specific rule
- `combine` will use the `Merge()` method of the `Option`

The merge strategy can be configured on ruleset level:
``` yaml
providers:
- id: managedk8s
  name: "Managed Kubernetes"
  rulesets:
  - id: disa-kubernetes-stig
    name: DISA Kubernetes Security Technical Implementation Guide
    version: v2r4
    args:
      mergeStrategy: replace
```

**Why is this needed**:
Enhance the user experience of writing diki configs. Provide a way without editing recommended rule options to add on additional.

AleksandarSavchev:
Merge options have been implemented with the referenced PRs.

PR #705, opened 2026-04-21
Add support for mergeable rules

AleksandarSavchev:

/kind enhancement

This PR adds a `diki config merge` command that combines a base (default) configuration with a current configuration. The base config defines the default rule options for an environment, while the current config provides user-specific overrides and runtime args. Only `ruleOptions` are merged — the current config is primary for everything else (provider args, metadata, output settings).

The command validates the current config before merging and the merged output after merging.

Usage:
```bash
diki config merge --base base.yaml --current current.yaml --output merged.yaml
```

Part of #687 

**Release note**:

```feature user
Added `diki config merge` command that combines a base (default) configuration with a current configuration.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,21 @@ Diki can validate the structure and content of a configuration file.
 diki config validate config.yaml
 ```
 
+### Config Merge
+
+Diki can merge two configuration files, a base (default) config and a custom config, into a single config.
+This is helpful when you want to reuse the same rule options across different diki runs.
+Only `ruleOptions` are merged, everything else (provider args, metadata, output) comes from the custom config.
+
+If either config skips a rule, the merged result will always skip it.
+
+```bash
+diki config merge \
+    --base=base.yaml \
+    --custom=custom.yaml \
+    --output=merged.yaml
+```
+
 ### Unit Tests
 
 You can manually run the tests via `make test`.
--- a/README.md
+++ b/README.md
@@ -177,8 +177,8 @@ Merge behavior for `ruleOptions`:
 ```bash
 diki config merge \
     --base=base.yaml \
-    --output=merged.yaml \
-    current.yaml
+    --current=current.yaml \
+    --output=merged.yaml
 ```
 
 ### Unit Tests
--- a/README.md
+++ b/README.md
@@ -172,13 +172,13 @@ Merge behavior for `ruleOptions`:
 - Rules only in the current config are kept as-is
 - Rules only in the base config are appended to the output
 - Rules in both configs are merged if a merge strategy is registered, otherwise the current config wins
-- If either config skips a rule, the merged result will always skip it
+- If either config skips a rule, the merged result will skip it. If both configs skip a rule, the current config's justification takes priority
 
 ```bash
 diki config merge \
     --base=base.yaml \
-    --current=current.yaml \
-    --output=merged.yaml
+    --output=merged.yaml \
+    current.yaml
 ```
 
 ### Unit Tests
--- a/README.md
+++ b/README.md
@@ -164,16 +164,20 @@ diki config validate config.yaml
 
 ### Config Merge
 
-Diki can merge two configuration files, a base (default) config and a custom config, into a single config.
+Diki can merge two configuration files, a base (default) config and a current config, into a single config.
 This is helpful when you want to reuse the same rule options across different diki runs.
-Only `ruleOptions` are merged, everything else (provider args, metadata, output) comes from the custom config.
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -162,6 +162,21 @@ Diki can validate the structure and content of a configuration file.
 diki config validate config.yaml
 ```
 
+### Config Merge
+
+Diki can merge two configuration files, a base (default) config and a custom config, into a single config.
```
dimityrmirchev: Please document how the two configs are treated, e.g. if they are both validated or if there is a difference in which is the base and which is the other config.

cmd/diki/app/app.go
```diff
@@ -178,6 +179,13 @@ e.g. to check compliance of your hyperscaler accounts.`,
 		}
 	}
 
+	mergeRegistryFuncs := []provider.MergeRegistryFunc{}
```
dimityrmirchev:
```suggestion
var	mergeRegistryFuncs  []provider.MergeRegistryFunc
```

cmd/diki/app/app.go
```diff
@@ -203,6 +211,19 @@ e.g. to check compliance of your hyperscaler accounts.`,
 
 	configCmd.AddCommand(configValidateCmd)
 
+	var configMergeOpts configMergeOptions
+	configMergeCmd := &cobra.Command{
+		Use:   "merge",
+		Short: "Merge a base config with a custom config.",
```
dimityrmirchev: We should think of another way of calling the custom config

cmd/diki/main.go
```diff
@@ -10,25 +10,36 @@ import (
 	controllerruntime "sigs.k8s.io/controller-runtime"
 
 	"github.com/gardener/diki/cmd/diki/app"
+	"github.com/gardener/diki/pkg/config/merge"
 	"github.com/gardener/diki/pkg/provider"
 	"github.com/gardener/diki/pkg/provider/builder"
 	"github.com/gardener/diki/pkg/provider/garden"
+	gardenmerge "github.com/gardener/diki/pkg/provider/garden/ruleset/securityhardenedshoot"
 	"github.com/gardener/diki/pkg/provider/gardener"
+	gardenermerge "github.com/gardener/diki/pkg/provider/gardener/ruleset/disak8sstig"
```
dimityrmirchev: We should not add merge to the name of these imports

cmd/diki/main.go
```diff
@@ -10,25 +10,36 @@ import (
...
+			virtualgarden.ProviderID: {ProviderFromConfigFunc: builder.VirtualGardenProviderFromConfig, MetadataFunc: builder.VirtualGardenProviderMetadata, ValidateConfigFunc: virtualgarden.ValidateProviderConfig, MergeRegistryFunc: virtualgardenmerge.RegisterMergeFuncs},
 		},
 	)
 
 	if err := cmd.ExecuteContext(controllerruntime.SetupSignalHandler()); err != nil {
 		log.Fatal(err)
 	}
 }
+
+func managedk8sMergeRegistryFunc(r *merge.Registry) {
```
dimityrmirchev: Can we refactor this so we do not have to define a function when there are two or more functions?

pkg/config/merge/merge.go
```diff
@@ -0,0 +1,179 @@
...
+//   - Rule in both + MergeableOption: Merge() is called (base.Merge(custom))
+//   - Rule in both + not MergeableOption: custom args win
+//   - Skip always wins: if either config skips a rule, the merged result skips it
+func MergeConfigs(base, custom *config.DikiConfig, registry *Registry) (*config.DikiConfig, error) {
+	if base == nil {
+		return custom, nil
+	}
+	if custom == nil {
+		return custom, nil
+	}
```
dimityrmirchev:
```suggestion
	if custom == nil {
		return nil, nil
	}
```

dimityrmirchev (commented):
I am leaving a few preliminary comments.

cmd/diki/app/app.go
AleksandarSavchev: I have now renamed it to `current`, as to note that this is the config that would take priority and would work as a counter word for base.

We can also think of another name, we can also just pass the "custom" config without a flag.

dimityrmirchev: I like the "current" name. Can we also do `diki config merge base.yaml current.yaml --output foo.yaml`? Or do you want to keep the flags?

dimityrmirchev (commented):
/lgtm
/approve

PR #736, opened 2026-06-11
chore(ci): use client-id for gardener-github-pkg-mngr app token

petersutter:
/kind cleanup

The `app-id` input of `actions/create-github-app-token` is deprecated in favor of `client-id`. This PR renames the workflow secret from `GARDENER_GITHUB_WORKFLOW_PKG_MNGR_APP_ID` to `GARDENER_GITHUB_WORKFLOW_PKG_MNGR_CLIENT_ID` and passes it as `client-id` to the action.

Fixes #

The new secret `GARDENER_GITHUB_WORKFLOW_PKG_MNGR_CLIENT_ID` is already in place at the org level and ready to be consumed by the repos.

**Release note**:
```other dependency
NONE
```


dimityrmirchev (commented):
/lgtm
/approve

PR #746, opened 2026-06-23
[Security Hardened Shoot Cluster] Implement `MergeableOption` for all rules

AleksandarSavchev:

/kind enhancement

This PR implements the `MergeableOption` interface for all rules in the Security Hardened Shoot Cluster guide.

Part of https://github.com/gardener/diki/issues/687

**Release note**:

```feature user
[Security Hardened Shoot Cluster] All rules can now me merged using the `config merge` command.
```

### DIFF
--- a/pkg/provider/garden/ruleset/securityhardenedshoot/rules/1000.go
+++ b/pkg/provider/garden/ruleset/securityhardenedshoot/rules/1000.go
@@ -20,9 +20,10 @@ import (
 )
 
 var (
-	_ rule.Rule     = &Rule1000{}
-	_ rule.Severity = &Rule1000{}
-	_ option.Option = &Options1000{}
+	_ rule.Rule              = &Rule1000{}
+	_ rule.Severity          = &Rule1000{}
+	_ option.Option          = &Options1000{}
+	_ option.MergeableOption = &Options1000{}
 )
 
 type Options1000 struct {
@@ -31,6 +32,25 @@ type Options1000 struct {
 
 type Extension struct {
 	Type string `json:"type" yaml:"type"`
+}
+
+func (o *Options1000) Merge(other option.MergeableOption) (option.MergeableOption, error) {
+	if other == nil {
+		return o, nil
+	}
+
+	otherOpts, ok := other.(*Options1000)
+	if !ok {
+		return nil, fmt.Errorf("cannot merge options of type %T into *Options1000", other)
+	}
+
+	merged := &Options1000{
+		Extensions: make([]Extension, 0, len(o.Extensions)+len(otherOpts.Extensions)),
+	}
+	merged.Extensions = append(merged.Extensions, o.Extensions...)
+	merged.Extensions = append(merged.Extensions, otherOpts.Extensions...)
+
+	return merged, nil
 }
 
 func (o Options1000) Validate(fldPath *field.Path) field.ErrorList {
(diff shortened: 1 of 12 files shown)


pkg/provider/garden/ruleset/securityhardenedshoot/rules/2007.go
```diff
@@ -25,15 +25,37 @@ import (
...
+	if other == nil {
+		return o, nil
+	}
+
+	otherOpts, ok := other.(*Options2007)
+	if !ok {
+		return nil, fmt.Errorf("cannot merge options of type %T into *Options2007", other)
+	}
+
+	merged := &Options2007{
```
georgibaltiev, line 48: I am wondering if we shouldn't select the PSS Profile which gives more restrictions, since there is a possiblity that the user wants to configure something more restrictive - for example `Restricted`, while the default configuration specifies `Baseline`.

georgibaltiev (commented):
Thank you! The rest looks good, and I have only one question.

pkg/provider/garden/ruleset/securityhardenedshoot/rules/2007.go
AleksandarSavchev: I think the question is how would we want to handle such cases in general, when the options cannot be concatenated? I think on a merge the right hand side options should override the initial options.

This would work the same as in the scenario in `MergeConfigs()` where we have base and current options but without the `Merge()` method:
https://github.com/gardener/diki/blob/52b758b3e65dfe2212bd9aeb035dfa1774331347/pkg/config/merge/merge.go#L15-L20

georgibaltiev: Okay, that is consistent with the current approach.

pkg/provider/garden/ruleset/securityhardenedshoot/rules/1000.go
```diff
@@ -33,6 +34,25 @@ type Extension struct {
...
+	otherOpts, ok := other.(*Options1000)
+	if !ok {
+		return nil, fmt.Errorf("cannot merge options of type %T into *Options1000", other)
+	}
+
+	merged := &Options1000{
+		Extensions: make([]Extension, 0, len(o.Extensions)+len(otherOpts.Extensions)),
+	}
+	merged.Extensions = append(merged.Extensions, o.Extensions...)
+	merged.Extensions = append(merged.Extensions, otherOpts.Extensions...)
```
dimityrmirchev, line 51: Do we need to deduplicate here or anywhere else where we append configs?

georgibaltiev (commented):
/lgtm

dimityrmirchev (commented):
Just one question

pkg/provider/garden/ruleset/securityhardenedshoot/rules/1000.go
AleksandarSavchev: I had decided to leave slices of structs without deduplicating them, since we would need individual handling for different structs.

In another pr https://github.com/gardener/diki/pull/748 I added a new utils func for string slices https://github.com/AleksandarSavchev/diki/blob/2b64e504208b35d5d9f19480c26d1895e0332a3f/pkg/internal/utils/set.go#L71-L74

PR #748, opened 2026-06-25
[DISA K8s STIG] Implement `MergeableOption` for rule options in the `gardener` provider

AleksandarSavchev:

/kind enhancement

This PR implements the `MergeableOption` interface for all rules of the gardener provider in DISA K8s STIG.

Part of https://github.com/gardener/diki/issues/687

**Release note**:

```feature user
[DISA K8s STIG] All rules in the `gardener` provider can now be merged using the `config merge` command.
```

### DIFF
--- a/pkg/internal/utils/set.go
+++ b/pkg/internal/utils/set.go
@@ -6,6 +6,8 @@ package utils
 
 import (
 	"slices"
+
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // EqualSets checks if two slices contain exactly the same elements independent of the ordering.
@@ -64,4 +66,9 @@ func MatchLabels(m1, m2 map[string]string) bool {
 	}
 
 	return true
+}
+
+// MergeStringSlices merges two string slices and removes duplicates.
+func MergeStringSlices(s1, s2 []string) []string {
+	return sets.New(s1...).Insert(s2...).UnsortedList()
 }
--- a/pkg/internal/utils/set.go
+++ b/pkg/internal/utils/set.go
@@ -6,8 +6,6 @@ package utils
 
 import (
 	"slices"
-
-	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // EqualSets checks if two slices contain exactly the same elements independent of the ordering.
@@ -68,7 +66,24 @@ func MatchLabels(m1, m2 map[string]string) bool {
 	return true
 }
 
-// MergeStringSlices merges two string slices and removes duplicates.
+// MergeStringSlices merges two string slices and removes duplicates, preserving the order of s1 followed by s2.
 func MergeStringSlices(s1, s2 []string) []string {
-	return sets.New(s1...).Insert(s2...).UnsortedList()
+	var (
+		result []string
+		seen   = map[string]struct{}{}
+	)
+
+	for _, s := range s1 {
+		if _, ok := seen[s]; !ok {
+			seen[s] = struct{}{}
+			result = append(result, s)
+		}
+	}
+	for _, s := range s2 {
+		if _, ok := seen[s]; !ok {
+			seen[s] = struct{}{}
+			result = append(result, s)
+		}
+	}
+	return result
 }
--- a/pkg/provider/gardener/ruleset/disak8sstig/rules/242400.go
+++ b/pkg/provider/gardener/ruleset/disak8sstig/rules/242400.go
@@ -35,8 +35,10 @@ import (
 )
 
 var (
-	_ rule.Rule     = &Rule242400{}
-	_ rule.Severity = &Rule242400{}
+	_ rule.Rule              = &Rule242400{}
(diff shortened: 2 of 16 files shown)


pkg/provider/gardener/ruleset/disak8sstig/rules/242400.go
```diff
@@ -54,7 +56,20 @@ type Options242400 struct {
 	KubeProxy disaoption.KubeProxyOptionsWithoutSelectors `json:"kubeProxy" yaml:"kubeProxy"`
 }
 
-var _ option.Option = (*Options242400)(nil)
+func (o *Options242400) Merge(other option.MergeableOption) (option.MergeableOption, error) {
```
georgibaltiev, line 59: Since there is a few rules which only have a `KubeProxyOptionsWithoutSelectors`, wouldn't it be better, if we implement the merging logic in the `shared/ruleset/disak8sstig/option/options.go` file?

pkg/provider/gardener/ruleset/disak8sstig/rules/242400.go
```diff
@@ -54,7 +56,20 @@ type Options242400 struct {
...
+		return o, nil
+	}
+
+	otherOpts, ok := other.(*Options242400)
+	if !ok {
+		return nil, fmt.Errorf("cannot merge options of type %T into *Options242400", other)
+	}
+
+	return &Options242400{
+		KubeProxy: otherOpts.KubeProxy,
```
georgibaltiev, line 70: What if the otherOpts.KubeProxy field is not set as well?

pkg/internal/utils/set.go
```diff
@@ -65,3 +67,8 @@ func MatchLabels(m1, m2 map[string]string) bool {
 
 	return true
 }
+
+// MergeStringSlices merges two string slices and removes duplicates.
+func MergeStringSlices(s1, s2 []string) []string {
+	return sets.New(s1...).Insert(s2...).UnsortedList()
```
georgibaltiev: By creating an unsorted list, is there a concern that running this function with the same inputs will produce different results? This may not be relevant at all, but still worth noting imo.

pkg/shared/ruleset/disak8sstig/option/options.go
georgibaltiev, line 1: Additionally, would we want to document the different ways rules merge, since a few of them are a simple concatenation, others are non-list values and third ones create a union without any duplicate values?

georgibaltiev (changes requested):
Thank you! I have some initial comments

pkg/provider/gardener/ruleset/disak8sstig/rules/242400.go
AleksandarSavchev: Currently `KubeProxyOptionsWithoutSelectors` contains only a single field of type `bool`. The merging logic is only to use the `other` field,  I do not think that would be needed.

pkg/provider/gardener/ruleset/disak8sstig/rules/242400.go
AleksandarSavchev: If the options are not `nil` the `KubeProxy` field would always be set, since it is not a pointer.

pkg/internal/utils/set.go
AleksandarSavchev: Sure, I can see it being a problem if we have future cases where order is important.

pkg/shared/ruleset/disak8sstig/option/options.go
AleksandarSavchev: The comment on [MergeConfig()](https://github.com/gardener/diki/blob/52b758b3e65dfe2212bd9aeb035dfa1774331347/pkg/config/merge/merge.go#L15-L20) and the [README](https://github.com/gardener/diki#config-merge) part should be used as reference:

The other (current) config takes priority when options cannot be concatenated, the deduplication is simply an optimization for concat cases.

georgibaltiev: Okay, I agree with that one.

pkg/provider/gardener/ruleset/disak8sstig/rules/242400.go
georgibaltiev: I agree.

pkg/provider/gardener/ruleset/disak8sstig/rules/242400.go
georgibaltiev: Sure, but if in the future these options or the merging logic require changes we would have to modify code in multiple places. I am okay with both approaches, just believe that it will be cleaner like that.

AleksandarSavchev: Yes, we will add the `Merge()` method when required.

dimityrmirchev (commented):
/lgtm
/approve

Issue #754, opened 2026-06-30
Implement `Pod Security Standards` ruleset


AleksandarSavchev:
**What would you like to be added**:

A new `Pod Security Standards` ruleset should be implemented for the `managedK8s` provider. This ruleset checks pods running in a managed Kubernetes cluster against the [Kubernetes Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/). The ruleset should implement the Baseline and Restricted profiles as described in the official Kubernetes documentation.

### Baseline Profile

The Baseline profile prevents known privilege escalations and covers the following controls:
- HostProcess (Windows)
- Host Namespaces (hostNetwork, hostPID, hostIPC)
- Privileged Containers
- Capabilities (restrict to safe set)
- HostPath Volumes
- Host Ports
- AppArmor
- SELinux
- /proc Mount Type
- Seccomp
- Sysctls

### Restricted Profile

The Restricted profile enforces current Pod hardening best practices and includes all Baseline controls plus:
- Volume Types (only configMap, csi, downwardAPI, emptyDir, ephemeral, persistentVolumeClaim, projected, secret allowed)
- Privilege Escalation (allowPrivilegeEscalation must be false)
- Running as Non-Root (runAsNonRoot must be true)
- Running as Non-Root User (runAsUser must not be 0)

**Why is this needed**:

Pod Security Standards are the official Kubernetes mechanism for defining security constraints at the pod level. Implementing them as a diki ruleset enables automated compliance checking for managed Kubernetes clusters, covering common security observations such as missing `readOnlyRootFilesystem`, unrestricted `allowPrivilegeEscalation`, missing capabilities restrictions, `runAsNonRoot` not enforced, and use of `privileged`, `hostNetwork`, or `hostPID`.

PR #755, opened 2026-07-02
[Pod Security Standards] Add ruleset documentation

AleksandarSavchev:

/kind enhancement

Adds documentation for the new Pod Security Standards ruleset. This ruleset will check pods running in a managed Kubernetes cluster against the official [Kubernetes Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/), covering both the Baseline (PSS-B001 to PSS-B012) and Restricted (PSS-R001 to PSS-R006) profiles.

Part of https://github.com/gardener/diki/issues/754

**Release note**:

```doc user
Add documentation for the Pod Security Standards ruleset.
```

### DIFF
--- a/docs/rulesets/pod-security-standards/pod-security-standards-v0.1.0.yaml
+++ b/docs/rulesets/pod-security-standards/pod-security-standards-v0.1.0.yaml
@@ -1,0 +1,83 @@
+# SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+ruleset:
+  id: pod-security-standards
+  name: "Pod Security Standards"
+  version: "v0.1.0"
+rules:
+# Baseline Profile
+- id: PSS-B001
+  name: "HostProcess"
+  description: "Windows Pods offer the ability to run HostProcess containers which enables privileged access to the Windows host machine. Privileged access to the host is disallowed in the Baseline policy."
+  severity: "HIGH"
+- id: PSS-B002
+  name: "Host Namespaces"
+  description: "Sharing the host namespaces must be disallowed."
+  severity: "HIGH"
+- id: PSS-B003
+  name: "Privileged Containers"
+  description: "Privileged Pods disable most security mechanisms and must be disallowed."
+  severity: "HIGH"
+- id: PSS-B004
+  name: "Capabilities (Baseline)"
+  description: "Adding additional capabilities beyond those listed below must be disallowed."
+  severity: "HIGH"
+- id: PSS-B005
+  name: "HostPath Volumes"
+  description: "HostPath volumes must be forbidden."
+  severity: "HIGH"
+- id: PSS-B006
+  name: "Host Ports"
+  description: "HostPorts should be disallowed entirely (recommended) or restricted to a known list."
+  severity: "HIGH"
+- id: PSS-B007
+  name: "Host Probes / Lifecycle Hooks"
+  description: "The host field in probes and lifecycle hooks must be disallowed."
+  severity: "HIGH"
+- id: PSS-B008
+  name: "AppArmor"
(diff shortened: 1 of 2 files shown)


docs/rulesets/pod-security-standards/pod-security-standards-v0.1.0.yaml
```diff
@@ -0,0 +1,83 @@
...
+  name: "Host Namespaces"
+  description: "Sharing the host namespaces must be disallowed."
+  severity: "HIGH"
+- id: PSS-B003
+  name: "Privileged Containers"
+  description: "Privileged Pods disable most security mechanisms and must be disallowed."
+  severity: "HIGH"
+- id: PSS-B004
+  name: "Capabilities (Baseline)"
+  description: "Adding additional capabilities beyond those listed below must be disallowed."
```
dimityrmirchev: we should probably list the capabilities?

docs/rulesets/pod-security-standards/pod-security-standards-v0.1.0.yaml
```diff
@@ -0,0 +1,83 @@
...
+  severity: "HIGH"
+- id: PSS-B005
+  name: "HostPath Volumes"
```
dimityrmirchev: Either that or we should adapt other similar names to be consistent across naming. Or was this taken from the k8s documentation?
```suggestion
  name: "Host Path Volumes"
```

docs/rulesets/pod-security-standards/pod-security-standards-v0.1.0.yaml
```diff
@@ -0,0 +1,83 @@
...
+  description: "Containers must not set runAsUser to 0."
+  severity: "MEDIUM"
+- id: PSS-R005
+  name: "Seccomp (Restricted)"
+  description: "Seccomp profile must be explicitly set to RuntimeDefault or Localhost. Unconfined profiles and the absence of a profile are prohibited."
+  severity: "MEDIUM"
+- id: PSS-R006
+  name: "Capabilities (Restricted)"
+  description: "Containers must drop ALL capabilities, and are only permitted to add back NET_BIND_SERVICE."
+  severity: "MEDIUM"
```
dimityrmirchev, line 83: I cannot decide if those should be MEDIUM or HIGH. Privilege escalation and runnning as non-root seem like a HIGH requirements to me. How can we decide what to change? The other options is to set everything to high?

docs/rulesets/pod-security-standards/pod-security-standards-v0.1.0.yaml
AleksandarSavchev: Yes, I have now added the list

The description field were taken directly from the k8s PSP documentation from the `policy` field, but specific one cannot be left without the list.

docs/rulesets/pod-security-standards/pod-security-standards-v0.1.0.yaml
AleksandarSavchev: These names were take directly from the k8s PSS documentation, as the name of the policy. I agree that there is an inconsistency, but should we fix it or stick to the way they are written in k8s?

docs/rulesets/pod-security-standards/pod-security-standards-v0.1.0.yaml
AleksandarSavchev: Some of these rules have more restrictive versions in the `baseline` policies. Such are: `PSS-R001` < `PSS-B005`; `PSS-R005` < `PSS-B011`; `PSS-R006` < `PSS-B004`.

We can also say that `PSS-R004` < `PSS-R003`.

I agree that we can set `PSS-R002` & `PSS-R003` to HIGH

dimityrmirchev (commented):
/lgtm
/approve

Issue #752, opened 2026-06-29
Support integration with The Gardenlinux Project's testing framework


georgibaltiev:
**What would you like to be added**:

[The Gardenlinux Project](https://github.com/gardenlinux/) has it’s own Python-based testing framework, which could evaluate compliance with different technical implementation guides, most notably - the [DISA's General Purpouse OS System Security Requirements Guide](https://cyber.trackr.live/stig/General_Purpose_Operating_System_Security_Requirements_Guide/3/3), implemented via [a set of PyTest rules](https://github.com/gardenlinux/gardenlinux/tree/main/tests/integration/security/compliance).

Currently, the organization ships the testing framework also as a [container image](https://github.com/gardenlinux/gardenlinux/pkgs/container/test), which could be used to evaluate the compliance of Kubernetes clusters, as described in [this documentation](https://github.com/gardenlinux/gardenlinux/pkgs/container/test). The testing framework also presents the option to export the PyTest results as a JUnit XML report.

Diki's testing framework is familiar with running Kubernetes Pods, intended to evaluate compliance on K8s Nodes - [example](https://github.com/gardener/diki/blob/main/pkg/provider/managedk8s/ruleset/disak8sstig/rules/242466.go). This allows for the integration with the aformentioned framework.

A new ruleset, called `gardenlinux` for the `managedk8s` provider could be added. The maintenance of the ruleset will remain within the realm of the Gardenlinux project. [An issue, suggesting a PoC was already opened](https://github.com/gardener/diki/issues/641), highlighting the same approach.

The integration of the testing framework envisions the following steps, which could be described as a contract:
- Diki deploys a Pod resource with the aforementioned image, specifically catered to run the DISA ruleset.
- The testing image evaluates the cluster environment and generates a JUnit XML report, which will be retrieved by diki via a `kubectl exec`-style command
- Diki parses the retrieved XML report via [the standard XML Golang package](https://pkg.go.dev/encoding/xml). A parser utility will be implemented here.
- The parsed report is converted to a JSON-style diki-report. Thereafter, an HTML report could be generated via the already existing CLI.

A few more anecdotes:
- Initially, only [DISA's General Purpouse OS System Security Requirements Guide](https://github.com/gardenlinux/gardenlinux/tree/main/tests/integration/security/compliance) will be supported
- Running individual rules via the `--rule-id` flag of the `run` command will not be supported for this ruleset. The reason is due to the external implementation of the ruleset.
[...]

PR #760, opened 2026-07-07
Introduce the Gardenlinux Testing Framework ruleset

georgibaltiev:

/kind enhancement

This PR introduces support for the `gardenlinux` ruleset, implemented explictly for the `managedk8s` provider. The PR also introduces a validation for the ruleset configuration, as well as configurable arguments, such as a label selector for Nodes, which will be used in a follow-up PR to select nodes for evaluation by the ruleset.

Part of #752 

An actual version of the gardenlinux testing image will be introduced with a follow-up PR, which will replace the current dummy version.
Documentation will be added in a separate PR

**Release note**:

```feature user
A new ruleset, called `gardenlinux`, based on the Gardenlinux Testing Framework has been registered for the `managedk8s` provider
```

### DIFF
--- a/pkg/provider/builder/managedk8s.go
+++ b/pkg/provider/builder/managedk8s.go
@@ -15,6 +15,7 @@ import (
 	"github.com/gardener/diki/pkg/provider"
 	"github.com/gardener/diki/pkg/provider/managedk8s"
 	"github.com/gardener/diki/pkg/provider/managedk8s/ruleset/disak8sstig"
+	"github.com/gardener/diki/pkg/provider/managedk8s/ruleset/gardenlinux"
 	"github.com/gardener/diki/pkg/provider/managedk8s/ruleset/securityhardenedk8s"
 	"github.com/gardener/diki/pkg/ruleset"
 )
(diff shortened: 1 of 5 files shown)


pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
```diff
@@ -0,0 +1,157 @@
...
+	// Versions are sorted from newest to oldest.
+	// TODO(georgibaltiev): introduce support for actual gardenlinux versions and remove dummy values
+	SupportedVersions = []string{"v0.1.0"}
+)
+
+// Ruleset implements the Gardenlinux Testing Framework ruleset.
+type Ruleset struct {
+	version    string
+	Config     *rest.Config
+	numWorkers int
```
AleksandarSavchev: Do we want to limit the creation of concurrent pods?

Since they are on different nodes they should not be a problem for the load. Also there are diki rules in the disa ruleset, which create a pod on all nodes and we do not limit those.

pkg/provider/builder/managedk8s.go
AleksandarSavchev, line 109: We should add the new ruleset here as well.

AleksandarSavchev (changes requested):
Thanks, overall looks good. Just one question and one nit.

pkg/provider/builder/managedk8s.go
georgibaltiev: I assumed that we wouldn't want to add it to the default configuration, since it's maintained externally, but I could add it as well. WDYT?

georgibaltiev: I've discussed it @AleksandarSavchev and we have reached a conclusion to add support in the default configuration once the ruleset has become productive.

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
georgibaltiev: I agree with the remark, I will drop the numWorkers field in favor of better parallelism.

PR #766, opened 2026-07-10
Fix release publish to package repos

AleksandarSavchev:

/kind bug

The `publish-to-package-repositories` used `component-descriptor` as input to get the digests for the diki binaries, however this output is no longer provided with https://github.com/gardener/cc-utils/pull/1642.

This PR uses the release version to compute the digests from the release itself.

Fixes #

This change has been validated with https://github.com/gardener/diki/actions/runs/29090982808/job/86355973552

**Release note**:

```bugfix developer
A bug causing release packages to not be sent to `homebrew` and `chocolatey` has been fixed.
```


AleksandarSavchev:
Looking into another solution ref: https://github.com/gardener/gardenctl-v2/pull/777

/hold

AleksandarSavchev:
I would prefer if we can use the solution provided here, rather than the one from https://github.com/gardener/gardenctl-v2/pull/777.
Since it would be easier to retriger a failed job and we would not rely on the `component-descriptor-artefact`

/unhold

dimityrmirchev (commented):
/lgtm
/approve

PR #767, opened 2026-07-15
Shell-escape dynamic input passed to Execute()

AleksandarSavchev:

/kind bug

Shell-escape dynamic values interpolated into commands passed to `podExecutor.Execute()` to prevent unintended shell metacharacter interpretation.

Fixes #

**Release note**:

```bugfix user
A bug causing incorrect shell command execution when dynamic input contains special characters has been fixed.
```


pkg/internal/utils/utils.go
```diff
@@ -60,6 +60,12 @@ func (fs FileStats) Dir() string {
 	return filepath.Dir(fs.Path)
 }
 
+// ShellEscape wraps s in single quotes, escaping any embedded single quotes.
+// This prevents shell metacharacter interpretation when s is interpolated into a shell command.
+func ShellEscape(s string) string {
+	return "'" + strings.ReplaceAll(strings.TrimSpace(s), "'", "'\\''") + "'"
```
georgibaltiev: I prefer for this function to refrain from additional trimming/cleanup of the string - call the Trim only where necessary.
```suggestion
	return "'" + strings.ReplaceAll(s, "'", "'\\''") + "'"
```

georgibaltiev (changes requested):
Thanks! I have one change request.

georgibaltiev (approved):
Thanks!
/lgtm
/approve

PR #772, opened 2026-07-16
Add a Pod deployment mechanism for the Gardenlinux Ruleset

georgibaltiev:

/kind enhancement

This PR is a follow-up to #760 

It expands the newly-introduced `gardenlinux` ruleset by defining a new Pod spec that runs the containerized testing framework, based on [their reccomendation](https://github.com/gardener/diki/pull/772#issue-4900554191) to run tests on K8s environments.

The Pod deploys two containers - the `gardenlinux` test container, and the `diki-ops` report-reader container, which has a sleep timeout of 15 minutes.

Once the testing container has finished (as checked by its' status) a `cat <path-to-report>` command is executed into the report-reader container, which retrieves the XML data as a singular string value.

The user can now specify which version of the `gardenlinux` tests could be ran, and as initial support versions [2298.0.0](https://github.com/gardenlinux/gardenlinux/pkgs/container/test/1035278466?tag=2298.0.0) and [2297.0.0](https://github.com/gardenlinux/gardenlinux/pkgs/container/test/1033038064?tag=2297.0.0) have been added. 

Part of #752 

The parsing of the XML reports will be done in a separate PR (as marked by the few TODOs left in ruleset.go)

**Release note**:

```feature user
[Gardenlinux Testing Framework] Added Pod resources to run the gardenlinux testing container.
```

### DIFF
--- a/pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod.go
+++ b/pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod.go
@@ -37,7 +37,7 @@ const (
 	// ReportMountPath is the in-pod path of the shared report volume.
 	ReportMountPath = "/tests/tests/output"
 	// ReportFilename is the name of the generated JUnit XML report inside the pod.
-	ReportFilename = "test.xml"
+	ReportFilename = "report.xml"
 	// ReportSizeLimitKi is the size limit of the report volume in kibibytes.
 	ReportSizeLimitKi = 500
 
@@ -74,6 +74,7 @@ func NewTestPod(name, namespace, gardenlinuxTestImage, reportReaderImage, nodeNa
 			Labels:    labels,
 		},
 		Spec: corev1.PodSpec{
+			ActiveDeadlineSeconds:        ptr.To[int64](300),
 			AutomountServiceAccountToken: ptr.To(false),
 			SecurityContext: &corev1.PodSecurityContext{
 				// The FSGroup ensures that the written XML report is group-owned by 65532 - the reader will be able to get the report even if the "other" permission bits are dropped by the testing framework in a future release.
@@ -97,7 +98,7 @@ func NewTestPod(name, namespace, gardenlinuxTestImage, reportReaderImage, nodeNa
 						"-m",
 						"security_id",
 						"--junit-xml",
-						"output/test.xml",
+						"output/report.xml",
 						"--system-booted",
 						"--expected-users",
 						"gardener",
@@ -117,7 +118,7 @@ func NewTestPod(name, namespace, gardenlinuxTestImage, reportReaderImage, nodeNa
 					},
 					Args: []string{
 						"sleep",
-						"900",
+						"60",
 					},
 					SecurityContext: &corev1.SecurityContext{
 						Privileged:               ptr.To(false),
--- a/pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod.go
+++ b/pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod.go
@@ -1,0 +1,188 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package pod
+
+import (
+	"maps"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kubernetespod "github.com/gardener/diki/pkg/kubernetes/pod"
+)
+
+const (
+	// SystemNamespace is the namespace in which the test pods are created.
+	SystemNamespace = "kube-system"
+
+	// PodContextWaitInterval is the interval between pod status polls.
+	PodContextWaitInterval = 2 * time.Second
+	// PodContextWaitTimeout is the maximum time to wait for the test pod to reach Running.
+	// Both containers start immediately, so we only need enough time for image pull + scheduling.
+	PodContextWaitTimeout = 10 * time.Minute
+
+	// TestContainerName is the name of the container running the gardenlinux/tests suite.
+	TestContainerName = "gardenlinux-test"
+	// ReaderContainerName is the name of the container from which diki execs to read the completed report.
+	// TODO (georgibaltiev): change the name of this container once https://github.com/gardener/diki/issues/771 has been addressed.
+	// The container name coincides with the targeted container in the SimplePodExecutor.
+	ReaderContainerName = "container"
+
+	// ReportVolumeName is the name of the shared volume holding the JUnit report.
+	ReportVolumeName = "test-report"
+	// ReportMountPath is the in-pod path of the shared report volume.
+	ReportMountPath = "/tests/tests/output"
+	// ReportFilename is the name of the generated JUnit XML report inside the pod.
+	ReportFilename = "test.xml"
+	// ReportSizeLimitKi is the size limit of the report volume in kibibytes.
+	ReportSizeLimitKi = 500
+
+	// HostRootVolumeName is the name of the volume mounting the host root filesystem.
+	HostRootVolumeName = "host-root-volume"
+	// HostRootMountPath is the in-pod path at which the host root filesystem is mounted.
+	HostRootMountPath = "/host"
+	// HostRootPath is the host path mounted into the reader container.
+	HostRootPath = "/"
+
+	// LabelComplianceRoleGardenlinuxTestPod is used as the label value for LabelComplianceRoleKey indicating gardenlinux test pods.
+	LabelComplianceRoleGardenlinuxTestPod = "gardenlinux-test-pod"
+)
+
+// NewPodContext creates a new pod.PodContext tuned for the long-running gardenlinux/tests container.
+func NewPodContext(c client.Client, config *rest.Config, additionalPodLabels map[string]string) (*kubernetespod.SimplePodContext, error) {
+	podContext, err := kubernetespod.NewSimplePodContext(c, config, additionalPodLabels)
+	if err != nil {
+		return nil, err
+	}
+	podContext.WaitInterval = PodContextWaitInterval
+	podContext.WaitTimeout = PodContextWaitTimeout
+	return podContext, nil
+}
+
+// NewTestPod creates a new gardenlinux-test Pod.
+func NewTestPod(name, gardenlinuxTestImage, reportReaderImage, nodeName string, additionalLabels map[string]string) func() *corev1.Pod {
+	if len(name) > kubernetespod.MaxPodNameLength {
+		name = name[:kubernetespod.MaxPodNameLength]
+	}
+
+	labels := map[string]string{}
+	if additionalLabels != nil {
+		labels = maps.Clone(additionalLabels)
+	}
+
+	testPod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: SystemNamespace,
+			Labels:    labels,
+		},
+		Spec: corev1.PodSpec{
+			AutomountServiceAccountToken: ptr.To(false),
+			SecurityContext: &corev1.PodSecurityContext{
+				// The FSGroup ensures that the written XML report is group-owned by 65532 - the reader will be able to get the report even if the "other" permission bits are dropped by the testing framework in a future release.
+				FSGroup: ptr.To(int64(65532)),
+				SeccompProfile: &corev1.SeccompProfile{
+					Type: corev1.SeccompProfileTypeRuntimeDefault,
+				},
+			},
+			Containers: []corev1.Container{
+				{
+					Name:  TestContainerName,
+					Image: gardenlinuxTestImage,
+					// The gardenlinux testing container requires certain privileged access, as described in the official documentation - https://github.com/gardenlinux/gardenlinux/tree/main/tests#gardener--kubernetes-cluster-live-tests
+					SecurityContext: &corev1.SecurityContext{
+						Privileged: ptr.To(true),
+					},
+					Args: []string{
+						"./run_tests",
+						"integration/security/compliance",
+						"-v",
+						"-m",
+						"security_id",
+						"--junit-xml",
+						"output/test.xml",
+						"--system-booted",
+						"--expected-users",
+						"gardener",
+					},
+					VolumeMounts: []corev1.VolumeMount{
+						{
+							Name:      ReportVolumeName,
+							MountPath: ReportMountPath,
+						},
+					},
+				},
+				{
+					Name:  ReaderContainerName,
+					Image: reportReaderImage,
+					Command: []string{
+						"/bin/busybox",
+					},
+					Args: []string{
+						"sleep",
+						"900",
+					},
+					SecurityContext: &corev1.SecurityContext{
+						Privileged:               ptr.To(false),
+						AllowPrivilegeEscalation: ptr.To(false),
+						RunAsNonRoot:             ptr.To(true),
+						RunAsUser:                ptr.To(int64(65532)),
+						RunAsGroup:               ptr.To(int64(65532)),
+						ReadOnlyRootFilesystem:   ptr.To(true),
+						Capabilities: &corev1.Capabilities{
+							Drop: []corev1.Capability{"ALL"},
+						},
+					},
+					VolumeMounts: []corev1.VolumeMount{
+						{
+							Name:      ReportVolumeName,
+							MountPath: ReportMountPath,
+							ReadOnly:  true,
+						},
+					},
+				},
+			},
+			HostNetwork:   true,
+			HostPID:       true,
+			RestartPolicy: corev1.RestartPolicyNever,
+			Tolerations: []corev1.Toleration{
+				{
+					Effect:   corev1.TaintEffectNoSchedule,
+					Operator: corev1.TolerationOpExists,
+				},
+				{
+					Effect:   corev1.TaintEffectNoExecute,
+					Operator: corev1.TolerationOpExists,
+				},
+			},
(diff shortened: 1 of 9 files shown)


pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
```diff
@@ -132,19 +154,135 @@ func ValidateRulesetConfig(rulesetConfig config.RulesetConfig, fldPath *field.Pa
...
+		return ruleset.RulesetResult{}, err
+	}
+
+	nodesAllocatablePods := kubeutils.GetNodesAllocatablePodsNum(allClusterPods, nodes)
+	selectedNodes, _ := kubeutils.SelectNodes(nodes, nodesAllocatablePods, r.args.NodeGroupByLabels)
+
+	var rulesetResults []ruleset.RulesetResult
+	for _, node := range selectedNodes {
+		result, err := r.runOnNode(ctx, node.Name, testImage.String(), sidecarImage.String())
+		if err != nil {
```
AleksandarSavchev: Let's use go routines to have the tests run in parallel here.

The load should not be a problem since each pod is created on a different node.

georgibaltiev: I've force-pushed the branch with the new implementation, in order to avoid messy diffs between the old and the new implementation.

pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod.go
```diff
@@ -0,0 +1,188 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kubernetespod "github.com/gardener/diki/pkg/kubernetes/pod"
+)
+
+const (
+	// SystemNamespace is the namespace in which the test pods are created.
+	SystemNamespace = "kube-system"
```
AleksandarSavchev: Can we pass the namespace to the `NewTestPod()` func?

pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// PodContextWaitInterval is the interval between pod status polls.
+	PodContextWaitInterval = 2 * time.Second
+	// PodContextWaitTimeout is the maximum time to wait for the test pod to reach Running.
+	// Both containers start immediately, so we only need enough time for image pull + scheduling.
+	PodContextWaitTimeout = 10 * time.Minute
```
AleksandarSavchev: This timeout might be too long. Waiting for 10 minutes before we fail.

We can change it to 1 minute (as is for the privileged pod) or maybe 2 minutes if we have concerns that the test image is too big and would take some time

pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod.go
```diff
@@ -0,0 +1,188 @@
...
+	ReportFilename = "test.xml"
+	// ReportSizeLimitKi is the size limit of the report volume in kibibytes.
+	ReportSizeLimitKi = 500
+
+	// HostRootVolumeName is the name of the volume mounting the host root filesystem.
+	HostRootVolumeName = "host-root-volume"
+	// HostRootMountPath is the in-pod path at which the host root filesystem is mounted.
+	HostRootMountPath = "/host"
+	// HostRootPath is the host path mounted into the reader container.
+	HostRootPath = "/"
```
AleksandarSavchev: Do we need to mount anything on the node?

pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod.go
```diff
@@ -0,0 +1,188 @@
...
+					VolumeMounts: []corev1.VolumeMount{
+						{
+							Name:      ReportVolumeName,
+							MountPath: ReportMountPath,
+							ReadOnly:  true,
+						},
+					},
+				},
+			},
+			HostNetwork:   true,
```
AleksandarSavchev: Is this required? I do not see it in the [docu](https://github.com/gardenlinux/gardenlinux/tree/main/tests#gardener--kubernetes-cluster-live-tests)

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
```diff
@@ -132,24 +161,161 @@ func ValidateRulesetConfig(rulesetConfig config.RulesetConfig, fldPath *field.Pa
...
+	if err != nil {
+		return ruleset.RulesetResult{}, fmt.Errorf("failed to find image version for %s: %w", images.DikiOpsImageName, err)
+	}
+
+	nodes, err := kubeutils.GetNodes(ctx, r.Client, 300)
+	if err != nil {
+		return ruleset.RulesetResult{}, err
+	}
+
+	allClusterPods, err := kubeutils.GetPods(ctx, r.Client, "", labels.NewSelector(), 300)
```
AleksandarSavchev:
```suggestion
	allPods, err := kubeutils.GetPods(ctx, r.Client, "", labels.NewSelector(), 300)
```

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
```diff
@@ -132,24 +161,161 @@ func ValidateRulesetConfig(rulesetConfig config.RulesetConfig, fldPath *field.Pa
...
+	if err != nil {
+		return ruleset.RulesetResult{}, err
+	}
+
+	nodesAllocatablePods := kubeutils.GetNodesAllocatablePodsNum(allClusterPods, nodes)
+	selectedNodes, _ := kubeutils.SelectNodes(nodes, nodesAllocatablePods, r.args.NodeGroupByLabels)
```
AleksandarSavchev, line 199: Add comment here that we need to address these checkResults/ warnings in the future

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
```diff
@@ -132,24 +161,161 @@ func ValidateRulesetConfig(rulesetConfig config.RulesetConfig, fldPath *field.Pa
...
+
+	// TODO (georgibaltiev): return a merged RulesetResult, based on the gathered results from the slices
+	if len(rulesetResults) == 0 {
+		return ruleset.RulesetResult{}, nil
+	}
+	return rulesetResults[0], nil
+}
+
+func (r *Ruleset) runOnNode(ctx context.Context, nodeName, testImage, sidecarImage string) (ruleset.RulesetResult, error) {
+	podName := fmt.Sprintf("test-%s-%s", r.ID(), sharedrules.Generator.Generate(10))
```
AleksandarSavchev:
```suggestion
	podName := fmt.Sprintf("diki-%s-%s", r.ID(), sharedrules.Generator.Generate(10))
```

I think this should be consistent with our other pod names. Also, since we are creating them in parallel there is a race condition with the `Generator.Generate(10)`, where we can pods with the same name

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
```diff
@@ -32,17 +44,18 @@ var (
...
 type Ruleset struct {
-	version    string
-	Config     *rest.Config
-	args       Args
-	instanceID string
-	logger     *slog.Logger
+	version           string
+	Config            *rest.Config
+	Client            client.Client
+	ClusterPodContext pod.SimplePodContext
```
AleksandarSavchev:
```suggestion
	PodContext pod.SimplePodContext
```

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
```diff
@@ -132,24 +161,161 @@ func ValidateRulesetConfig(rulesetConfig config.RulesetConfig, fldPath *field.Pa
...
+		close(resultCh)
+	}()
+
+	finishMsg := "finished ruleset run"
+	resultCount := 0
+	for rulesetRun := range resultCh {
+		resultCount++
+		remaining := len(selectedNodes) - resultCount
+		if rulesetRun.err != nil {
+			r.Logger().Error(finishMsg, "ruleset_id", RulesetID, "node_name", rulesetRun.nodeName, "remaining_nodes", remaining, "error", rulesetRun.err)
```
AleksandarSavchev:
```suggestion
			r.Logger().Error(finishMsg, "node_name", rulesetRun.nodeName, "remaining_nodes", remaining, "error", rulesetRun.err)
```

We already log the ruleset ID.

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
```diff
@@ -132,24 +161,161 @@ func ValidateRulesetConfig(rulesetConfig config.RulesetConfig, fldPath *field.Pa
...
+			err = errors.Join(err, fmt.Errorf("ruleset %s on node %s errored: %w", RulesetID, rulesetRun.nodeName, rulesetRun.err))
+		} else {
+			r.Logger().Info(finishMsg, "ruleset_id", RulesetID, "node_name", rulesetRun.nodeName, "remaining_nodes", remaining)
```
AleksandarSavchev:
```suggestion
			r.Logger().Info(finishMsg, "node_name", rulesetRun.nodeName, "remaining_nodes", remaining)
```

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
```diff
@@ -132,24 +161,161 @@ func ValidateRulesetConfig(rulesetConfig config.RulesetConfig, fldPath *field.Pa
...
+	if err := retry.Until(timeoutCtx, r.ClusterPodContext.WaitInterval, func(ctx context.Context) (bool, error) {
+		terminated, err := r.testContainerTerminated(ctx, podName)
+		if err != nil {
+			return retry.MinorError(err)
+		}
+		if !terminated {
+			return retry.MinorError(fmt.Errorf("gardenlinux test container %q has not terminated yet", gardenlinuxpod.TestContainerName))
+		}
+
+		report, err = podExecutor.Execute(ctx, "/bin/busybox cat /tests/tests/output/test.xml", "")
```
AleksandarSavchev: We should use the exported constants for `/tests/tests/output/test.xml`


```suggestion
		report, err = podExecutor.Execute(ctx, fmt.Sprintf("/bin/busybox cat %s/%s", gardenlinuxpod.ReportMountPath, gardenlinuxpod.ReportFilename), "")
```

pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+			Expect(p.Spec.AutomountServiceAccountToken).To(Equal(ptr.To(false)))
+			Expect(p.Spec.HostNetwork).To(BeTrue())
+			Expect(p.Spec.HostPID).To(BeTrue())
+			Expect(p.Spec.HostIPC).To(BeFalse())
+			Expect(p.Spec.RestartPolicy).To(Equal(corev1.RestartPolicyNever))
+			Expect(p.Spec.SecurityContext).NotTo(BeNil())
+			Expect(p.Spec.SecurityContext.FSGroup).To(Equal(ptr.To(int64(65532))))
+			Expect(p.Spec.SecurityContext.SeccompProfile).NotTo(BeNil())
+			Expect(p.Spec.SecurityContext.SeccompProfile.Type).To(Equal(corev1.SeccompProfileTypeRuntimeDefault))
```
AleksandarSavchev: These are a lot of expects, would it be easier if we have an expected pod spec and only change in for specific test cases?

AleksandarSavchev (changes requested):
Thanks, overall looks good. I might not have captured everything and will do a more thorough review after the suggestions are addresed.

pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod.go
georgibaltiev: You are right, this is a relique from an old PoC, where I was running a larget subset of framework rules and for non-gardenlinux nodes the timeout was a lot larger. running DISA STIG, this is no longer the case. I am making the timeout 1 minute.

georgibaltiev: Even without that, there are plans to avoid deploying the testing container on non-gardenlinux nodes.

pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod.go
georgibaltiev: I also forgot to remove this.

pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod.go
georgibaltiev: A leftover from testing, thanks for noticing!

pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod.go
georgibaltiev: Okay, I am hardcoding the constant into the function, similar to other invocations of NewPrivilegedPod
https://github.com/gardener/diki/blob/2203dec22b83d532ec731dc808edc75f74943a24/pkg/shared/ruleset/disak8sstig/rules/242404.go#L144

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
georgibaltiev: I'm just gonna extract the name generation before the goroutine gets "forked"

pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod_test.go
```diff
@@ -0,0 +1,224 @@
...
+						},
+						{
+							Name:  gardenlinuxpod.ReaderContainerName,
+							Image: sidecarImage,
+							Command: []string{
+								"/bin/busybox",
+							},
+							Args: []string{
+								"sleep",
+								"900",
```
AleksandarSavchev:
```suggestion
								"60",
```

We can have lower sleep time, since we expect diki to retrieve the report quickly

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
```diff
@@ -132,24 +161,161 @@ func ValidateRulesetConfig(rulesetConfig config.RulesetConfig, fldPath *field.Pa
...
+	if err != nil {
+		return ruleset.RulesetResult{}, fmt.Errorf("failed to find image version for %s: %w", images.GardenlinuxTestImageName, err)
+	}
+	testImage.WithOptionalTag(r.version)
+
+	sidecarImage, err := imagevector.ImageVector().FindImage(images.DikiOpsImageName)
+	if err != nil {
+		return ruleset.RulesetResult{}, fmt.Errorf("failed to find image version for %s: %w", images.DikiOpsImageName, err)
+	}
+
```
AleksandarSavchev, line 186:
```suggestion
		return ruleset.RulesetResult{}, fmt.Errorf("failed to find image version for %s: %w", images.DikiOpsImageName, err)
	}
    sidecarImage.WithOptionalTag(version.Get().GitVersion)

```

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
```diff
@@ -132,24 +161,161 @@ func ValidateRulesetConfig(rulesetConfig config.RulesetConfig, fldPath *field.Pa
...
+	if len(rulesetResults) == 0 {
+		return ruleset.RulesetResult{}, nil
+	}
+	return rulesetResults[0], nil
+}
+
+func (r *Ruleset) runOnNode(ctx context.Context, nodeName, podName, testImage, sidecarImage string) (ruleset.RulesetResult, error) {
+	defer func() {
+		timeoutCtx, cancel := context.WithTimeout(context.Background(), r.PodContext.WaitTimeout)
+		defer cancel()
```
AleksandarSavchev, line 259:
```suggestion
	defer func() {
		timeoutCtx, cancel := context.WithTimeout(context.Background(), time.Second*30)
		defer cancel()
```

Let's not reuse the wait timeout for pod to get ready here.

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
```diff
@@ -132,24 +161,161 @@ func ValidateRulesetConfig(rulesetConfig config.RulesetConfig, fldPath *field.Pa
...
+		return retry.Ok()
+	}); err != nil {
+		return "", fmt.Errorf("failed to read gardenlinux test report %s: %w", gardenlinuxpod.ReportFilename, err)
+	}
+
+	return report, nil
+}
+
+func (r *Ruleset) testContainerTerminated(ctx context.Context, podName string) (bool, error) {
+	testPod := &corev1.Pod{}
```
AleksandarSavchev: I think this function would be better suited for the gardenlinux `pod` package

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
AleksandarSavchev: We can add comment here that we will might want to allow the execution of unsupported version. We can also create an issue for this

AleksandarSavchev (changes requested):
Thanks, just a few nits.

pkg/provider/managedk8s/ruleset/gardenlinux/pod/pod_test.go
georgibaltiev: I agree, I've left it as a safeguard for investigation in case the Pod doesn't get deleted, but it will make more sense to complete the Pod faster.

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
georgibaltiev: I'm gonna ammend the original umbrella issue #752. I would like to follow-up a little bit later on this

georgibaltiev: For now I'll add a TODO

PR #779, opened 2026-07-21
Handle input version correctly for publish to package repos workflow

AleksandarSavchev:

/kind bug

A bug causing release packages to be sent with wrong version to `homebrew` and `chocolatey` has been fixed.

Fixes #

**Release note**:

```feature user
A bug causing release packages to be sent with wrong version to `homebrew` and `chocolatey` has been fixed.
```

### DIFF
--- a/.github/workflows/publish-to-package-repositories.yaml
+++ b/.github/workflows/publish-to-package-repositories.yaml
@@ -10,11 +10,11 @@ on:
       version:
         type: string
         required: true
-        description: The released version (semver without v-prefix, e.g. 0.x.0)
+        description: The released version (e.g. v0.x.0)
   workflow_dispatch:
     inputs:
       version:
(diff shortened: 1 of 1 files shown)


dimityrmirchev:
so https://github.com/gardener/diki/pull/766 is not actually fixing it?

AleksandarSavchev:
Without https://github.com/gardener/diki/pull/766 the workflow breaks and cannot be manually executed.

However it introduced a new bug, where the input that is passed from the `release` workflow DOES contain the `v` prefix and would use `vv` when sending to one of the package repos, ex: https://github.com/gardener/homebrew-tap/actions/runs/29826526045/job/88621012453

I triggered the [workflow manually](https://github.com/gardener/diki/actions/runs/29830115239) with version `v0.27.1` and we got the PR there https://github.com/gardener/homebrew-tap/pull/121

dimityrmirchev (commented):
/lgtm
/approve

PR #799, opened 2026-09-16
Introduce support for a LTR version of the Gardenlinux testing image

georgibaltiev:

/kind task

Until now, support for the Gardenlinux ruleset was via two unstable versions of the container testing image - 2324.0.0 and 2323.0.0. The newly-introduced stable backport of the container, starting with version 2150.10.1 is now available for use. 

Fixes #

**Release note**:

```feature user
Diki now supports stable version `2150.10.1` of the Gardenlinux testing framework.
```
```other user
Diki no longer supports experimental versions `2324.0.0` and `2323.0.0` of the Gardenlinux testing framework. 
```


AleksandarSavchev:
Thanks, please also update: https://github.com/gardener/diki/blob/0b25b10000b3c5b2ec4ad991e37b3f839319ef24/example/config/managedk8s.yaml#L323

We will release this today, so it should not be a problem.

PR #800, opened 2026-09-16
Add per-rule Skip options for the Gardenlinux ruleset

georgibaltiev:

/kind enhancement

Currently, the Gardenlinux ruleset, configured for the `Managed Kubernetes` provider forbids the users from setting per-rule options in the diki configuration file. This PR introduces partial support for such options, by allowing the user to specify a list of rules which could be skipped with an accompanying justification.

Since the ruleset is executed externally and diki can not keep track of the present `ruleID`s in the ruleset, skip options for rules with non-existent IDs are simply ignored.

Part of #752 

**Release note**:

```feature user
Diki now supports configurable skip options for rules of the Gardenlinux ruleset.
```


pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
```diff
@@ -153,8 +161,21 @@ func ValidateRulesetConfig(rulesetConfig config.RulesetConfig, fldPath *field.Pa
...
+		}
+
+		// Gardenlinux rules are executed by an external testing framework, so non-skip per-rule options are not accepted.
+		if ruleOption.Args != nil {
+			allErrs = append(allErrs, field.Forbidden(idxPath.Child("args"), "the gardenlinux ruleset does not accept per-rule args"))
+		}
+
+		if ruleOption.Skip != nil && ruleOption.Skip.Enabled && len(ruleOption.Skip.Justification) == 0 {
+			allErrs = append(allErrs, field.Required(idxPath.Child("skip", "justification"), "must not be empty when skip is enabled"))
+		}
```
AleksandarSavchev: Can we remove this check as we allow empty strings for Justification in our other rulesets. We can revisit this and decide if we want to change it in all places.

AleksandarSavchev (changes requested):
Thanks, one nit. Can you also add an example skip in the managedk8s.yaml example config?

pkg/provider/managedk8s/ruleset/gardenlinux/ruleset.go
georgibaltiev: I agree with the remark

AleksandarSavchev (approved):

